PR #1, opened 2018-11-29
Provides ArcLight configuration and Rake indexing tasks for EAD Documents

jrgriffiniii:
This blocks all current UI work to port the Princeton University Library branding, and is being actively tested against existing EAD-XML for indexing problems.


jrgriffiniii:
The commits within this pull request can be extracted into separate, more atomic pull requests which won't block progress elsewhere.

PR #2, opened 2018-11-30
Integrates support for deploying Capistrano

jrgriffiniii:
Please note that this will not work if the git repository being used is privately accessible (I have found a solution for this using [Personal API Tokens](https://capistranorb.com/documentation/getting-started/authentication-and-authorisation/) for this)

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -1,0 +1,39 @@
+# Load DSL and set up stages
+require "capistrano/setup"
+
+# Include default deployment tasks
+require "capistrano/deploy"
+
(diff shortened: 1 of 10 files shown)


jrgriffiniii:
I can look to close this out in favor of introducing more atomic PRs.

jrgriffiniii:
After a more thorough review, this doesn't appear to be more than the results of some minor customization of the configuration files following the invocation of `bundle exec cap install` for `Capistrano::Rails` (please see https://github.com/capistrano/rails/blob/master/README.md).

Issue #4, opened 2018-12-07
Apply the PUL branding


jrgriffiniii:
This involves applying the PUL branding using a combination of SCSS and JavaScript from other existing applications.  https://github.com/pulibrary/DSS/pull/90 was given as an example to follow (and introduces the possibility of including LUX in this process).

Some progress has been made on the branches https://github.com/pulibrary/pulfa3-light/tree/pul-branding-lux and https://github.com/pulibrary/pulfa3-light/tree/pul-branding.  `pul-branding` may no longer be needed given the advantages introduced by using Webpack to handle JavaScript assets.

jrgriffiniii:
Should this be closed before we request a review from Enterprise and User Services?

sdellis:
Blocked by #186

PR #32, opened 2019-10-18
Installs Webpacker and integrates the LUX design system for branding

jrgriffiniii:
Advances #4, by integrating Webpacker, Vue, and the LUX design system.  This also attempts to align the styling (at least, for the header) with that of Orangelight.


lib/tasks/plantain.rake
```diff
@@ -19,13 +19,13 @@ namespace :plantain do
   end
 
   desc "Run Solr and Arclight for interactive development"
-  task :development, %i[rails_server_args] do |_t, args|
+  task :development do
```
jrgriffiniii: I'm proposing this given that I kept encountering problems when trying to run the entire Solr/Rails stack with a single Rake Task in my environment (e. g. there would be false port collisions for the `solr_wrapper` Solr installation).

PR #58, opened 2019-11-05
Adjusts the Traject configuration for improved error handling and uses Sidekiq for batch indexing

hackmastera:


hackmastera:
I technically created this PR but it's really @jrgriffiniii's work; I just had to open this new one because the original was against a branch that has since been merged.

This PR fixes the index and makes it run in sidekiq. Merging for now; creating #59 to fix up rake tasks.

PR #112, opened 2020-02-11
finished all the required tests in ticket 84

### LINKED ISSUE #84: Evaluate and improve the testing for the EAD indexing using RSpec
The EAD indexing for Plantain isn't thoroughly tested, and providing more granular tests which ensure that the XML transformation is valid at the element level would be ideal.

Useful files:
 - [Traject config](https://github.com/pulibrary/plantain/blob/master/lib/plantain/traject/ead2_config.rb)
- [EAD XML fixture](https://github.com/pulibrary/plantain/blob/master/spec/fixtures/ead/mudd/publicpolicy/MC221.EAD.xml)
 - [Existing Traject rspec test](https://github.com/pulibrary/plantain/blob/master/spec/features/traject/ead2_indexing_spec.rb)

khanhvu98kk:
closes #84

### DIFF
--- a/spec/features/traject/ead2_indexing_spec.rb
+++ b/spec/features/traject/ead2_indexing_spec.rb
@@ -58,5 +58,56 @@ describe "EAD 2 traject indexing", type: :feature do
         )
       end
     end
+
+    it "gets the title tesim" do
+      expect(result["title_teim"]).to include(
+        "Harold B. Hoskins Papers"
+      )
+      expect(result["title_teim"]).to eq(
+        result["title_ssm"]
+      )
+    end
+
+    it "asserts that title filing si field is missing" do
+      expect(result["title_filing_si"]).to be_nil
+    end
+
+    context "YearRange normalizer tests" do
+      let(:years) { result["normalized_date_ssm"][0].split("-") }
+      let(:beginning) { years[0].to_i }
+      let(:ending) { years[1].to_i }
+
+      it "asserts YearRange normalizer works, that normalized_date_ssm contains start and end in date_range_sim field" do
+        expect(result["normalized_date_ssm"][0]).to include(
+          beginning.to_s,
+          ending.to_s
+        )
+      end
+
+      it "asserts YearRange normalizer works, the # of yrs in date_range_sim array correct, equal to difference between beginning and ending" do
+        expect(result["date_range_sim"].length).to equal(
+          ending - beginning + 1
+        )
+      end
+
+      it "asserts YearRange normalizer works, date_range_sim contains a random year between begin and end" do
+        expect(result["date_range_sim"]).to include(
+          rand(beginning..ending)
+        )
+      end
+    end
+
+    it "gets the normalized date" do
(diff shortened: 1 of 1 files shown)


spec/features/traject/ead2_indexing_spec.rb
```diff
@@ -58,5 +58,56 @@
...
+
+      it "asserts YearRange normalizer works, the # of yrs in date_range_sim array correct, equal to difference between beginning and ending" do
+        expect(result["date_range_sim"].length).to equal(
+          ending - beginning + 1
+        )
+      end
+
+      it "asserts YearRange normalizer works, date_range_sim contains a random year between begin and end" do
+        expect(result["date_range_sim"]).to include(
+          rand(beginning..ending)
```
jrgriffiniii: 👍

spec/features/traject/ead2_indexing_spec.rb
```diff
@@ -58,5 +58,56 @@
...
+    it "gets the title tesim" do
+      expect(result["title_teim"]).to include(
+        "Harold B. Hoskins Papers"
+      )
+      expect(result["title_teim"]).to eq(
+        result["title_ssm"]
+      )
+    end
+
+    it "asserts that title filing si field is missing" do
```
hackmastera: I think we should delete this test. If we have another record that does populate this field we can add a test for the field at that time.

spec/features/traject/ead2_indexing_spec.rb
```diff
@@ -58,5 +58,56 @@
         )
       end
     end
+
+    it "gets the title tesim" do
```
hackmastera: This tests for `title_teim` and `title_ssm`, but not `title_tesim`, as suggested by the name of the test. Do we want to add an assertion for `title_tesim`? Regardless, I think you could name the test "gets the title".

spec/features/traject/ead2_indexing_spec.rb
```diff
@@ -58,5 +58,56 @@
...
+      expect(result["title_teim"]).to eq(
+        result["title_ssm"]
+      )
+    end
+
+    it "asserts that title filing si field is missing" do
+      expect(result["title_filing_si"]).to be_nil
+    end
+
+    context "YearRange normalizer tests" do
```
hackmastera: Usually a context name starts with "when" so this could be "when there is a year range"

spec/features/traject/ead2_indexing_spec.rb
```diff
@@ -58,5 +58,56 @@
...
+      let(:years) { result["normalized_date_ssm"][0].split("-") }
+      let(:beginning) { years[0].to_i }
+      let(:ending) { years[1].to_i }
+
+      it "asserts YearRange normalizer works, that normalized_date_ssm contains start and end in date_range_sim field" do
```
hackmastera: Then the test name should be a continuation as if it's the rest of the sentence, like "normalizes the year range into start and end dates" -- the details of the precise field names, etc, are clear by looking at the body of the test.

PR #164, opened 2020-03-18
Add index service to index a single ead

hackmastera:
Used in development to get a bit of data in more quickly. Not sure
whether this is the desired architecture.

### DIFF
--- a/app/services/index_service.rb
+++ b/app/services/index_service.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class IndexService
+  # Index an EAD-XML Document into Solr
+  # @param [String] relative_path
+  def index_document(relative_path:, root_path: nil)
+    root_path ||= pulfa_root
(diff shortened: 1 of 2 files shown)


hackmastera:
Looks like this was extracted from the Rakefile. If the test can be made to pass we should replace the contents of that rake task with this service.

app/services/index_service.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class IndexService
+  # Index an EAD-XML Document into Solr
+  # @param [String] relative_path
+  def index_document(relative_path:, root_path: nil)
+    root_path ||= pulfa_root
```
jrgriffiniii: Is `pulfa_root` `nil` here?

hackmastera:
I believe progress here is blocked by #183 and the solr cleaning mentioned therein.

jrgriffiniii:
Deprecated by #256

PR #165, opened 2020-03-18
Add universal viewer via arclight's Viewer Class extension point

### LINKED ISSUE #35: Integrate the Universal Viewer
We will also need to be able to link to the manifests generated from Figgy. This is likely going to consist of just porting the existing viewer integration directly from Figgy itself.

hackmastera:
moves #35 to demo column


hackmastera:
i'll try reopening to see if that resolves ci issues.

PR #188, opened 2020-03-20
Fixes issues with seeding the Solr Collection with PULFA EADs and ensures that IndexJob commits to Solr

jrgriffiniii:
This should address cases where PULFA finding aid EADs are not being indexed properly with the necessary Rake tasks.

### DIFF
--- a/app/jobs/index_job.rb
+++ b/app/jobs/index_job.rb
@@ -59,6 +59,7 @@ class IndexJob < ApplicationJob
 
     logger.info("Requesting a batch Solr update...")
     blacklight_connection.add(solr_documents)
+    blacklight_connection.commit
     logger.info("Successfully indexed the EADs for #{file_paths.join(', ')}")
   end
 end
--- a/app/jobs/index_job.rb
+++ b/app/jobs/index_job.rb
@@ -59,7 +59,6 @@ class IndexJob < ApplicationJob
 
     logger.info("Requesting a batch Solr update...")
     blacklight_connection.add(solr_documents)
-    blacklight_connection.commit
     logger.info("Successfully indexed the EADs for #{file_paths.join(', ')}")
   end
 end
--- a/lib/pulfalight/traject/ead2_config.rb
+++ b/lib/pulfalight/traject/ead2_config.rb
@@ -62,9 +62,11 @@ end
 each_record do |_record, context|
   next unless settings["repository"]
 
-  context.clipboard[:repository] = Arclight::Repository.find_by(
+  repository = Arclight::Repository.find_by(
     slug: settings["repository"]
-  ).name
+  )
+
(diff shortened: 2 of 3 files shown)


lib/tasks/plantain.rake
```diff
@@ -66,18 +74,14 @@ namespace :plantain do
   desc "Seed fixture data to Solr"
   task :seed do
     puts "Seeding index with data from spec/fixtures/ead..."
-    Dir.glob("spec/fixtures/ead/*.xml").each do |file|
-      system("FILE=#{file} rake arclight:index") # no REPOSITORY_ID
-    end
 
-    Dir.glob("spec/fixtures/ead/*").each do |dir|
-      next unless File.directory?(dir)
+    Dir.glob("eads/**/*.xml").each do |file|
```
hackmastera: This task used to seed fixture data. Did we want it to index all eads instead? Do we need 2 tasks?

lib/plantain/traject/ead2_config.rb
```diff
@@ -62,9 +62,11 @@
 each_record do |_record, context|
   next unless settings["repository"]
 
-  context.clipboard[:repository] = Arclight::Repository.find_by(
```
hackmastera: I'm not sure what problem this change solves. Would it be relevant to add a test? I'm not sure what a `context.cllipboard` is.

app/jobs/index_job.rb
```diff
@@ -59,6 +59,7 @@ def perform(file_paths)
 
     logger.info("Requesting a batch Solr update...")
     blacklight_connection.add(solr_documents)
+    blacklight_connection.commit
```
hackmastera: In general we've moved away from committing as part of application code. We have soft and hard commits configured in the solrconfig (https://github.com/pulibrary/pul_solr/blob/352c072e0ab2e4913f39912d1704b1d8f93c9ab6/solr_configs/pulfalight-staging/conf/solrconfig.xml#L316-L325). Should we revisit this decision?

lib/tasks/plantain.rake
jrgriffiniii: I can separate this into two Rake tasks.

lib/plantain/traject/ead2_config.rb
jrgriffiniii: There were cases where `Arclight::Repository.find_by` returned a `nil`, and this triggered errors for indexing where the `Repository` had not been created.

app/jobs/index_job.rb
jrgriffiniii: No, I am not interested in moving away from the solrconfig-based approach. I was finding that I needed to `curl` in order to commit Documents which were indexed with Rake tasks, and I assumed that this was needed.

jrgriffiniii:
This is ready to be reviewed once again please.

Issue #162, opened 2020-03-18
Non-manifest digital object locations need to be indexed to a different field


hackmastera:
Figgy manifests and arks for objects in pudl are getting indexed to the same place. The result is that the viewer code tries to load arks into figgy viewers, resulting in a blank viewer and no link to the pudl object. We'll need to index them into different fields and display them differently. Perhaps until we get the design sorted out the viewer could be on the page and the link could go in the digital objects tab.

example title with a pudl ark: "1000 mile tree in the narrows"

hackmastera:
Alternately, should we index the "role" of the digital object location and check whether it is "https://iiif.io/api/presentation/2.1/" in the viewer / renderer logic?

sdellis:
This issue breaks any JavaScript on the page because it's trying to load insecure content via a secure page.

jrgriffiniii:
After discussing this on 03/23, we might need to confirm that the `@role` attribute of the `dao` element is acceptable in the EAD-parsing functionality for this work. We are looking to use `https://iiif.io/api/presentation/2.1/` as the value @regineheberlein

regineheberlein:
EAD2002, which is what we are currently using, uses xlink:role. The value is not controlled.

(EAD3 breaks the role attribute out into arcrole and linkrole, FWIW, though my understanding is that that is under review. https://www.loc.gov/ead/EAD3taglib/EAD3.html)

PR #191, opened 2020-03-24
Display digital objects as links if they aren't manifests

hackmastera:
refs #162 

looks like this now for non-manifest objects:

Screen Shot 2020-03-24 at 8 54 59 AM

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -16,5 +16,12 @@ module Plantain
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+
+    # load overrides
+    config.to_prepare do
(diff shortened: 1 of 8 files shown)


hackmastera:
Note: this will require a reindex to start working correctly.

config/application.rb
```diff
@@ -16,5 +16,12 @@ class Application < Rails::Application
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+
+    # load overrides
+    config.to_prepare do
+      Dir.glob(Rails.root.join("app", "**", "*_override*.rb")) do |c|
```
jrgriffiniii: 👍

Issue #170, opened 2020-03-18
"Collection context" is introducing performance problems


jrgriffiniii:
I'm finding that there are some expensive partial rendering calls which are blocking the rendering for component or collection Document Views. Specifically, the following partials seem to be rendered for every child Document within a given Document:

```
views/catalog/_index_collection_context_default.html.erb
views/catalog/_containers.html.erb
views/catalog/_online_content_label.html.erb (0.0ms)
views/catalog/_arclight_bookmark_control.html.erb (1.3ms)
views/catalog/_arclight_abstract_or_scope.html.erb (1.0ms)
views/catalog/_arclight_document_header_icon.html.erb
```

I suspect that this is introduced with an attempt to recursively generate the views for child and parent nodes in the finding aid Solr Document tree within the "Collection context" UI component.

jrgriffiniii:
Child recursion is now working for collections, with the contextual highlighting:

![image](https://user-images.githubusercontent.com/1443986/77516163-44dd1500-6e50-11ea-8200-507ebcaf9a1c.png)

jrgriffiniii:
The buttons for toggling the expansion/collapsing of child components have been implemented:

![image](https://user-images.githubusercontent.com/1443986/77543739-78369880-6e7e-11ea-8fd1-85ce93112bc0.png)

jrgriffiniii:
Currently I'm struggling to resolve the full structure of trees. Currently, I'm finding that file-level components are indexed with the following values for `parent_ssm`:

```
"parent_ssm":["ST1", "ST1_c000000001"],
```

Here both the immediate parent container, as well as the overall collection are indexed in the Document. However, attempting to query for all children of `ST1_c000000001` returned an empty set. Querying for all children of `ST1` returned file-level components but not `ST1_c000000001`.

Just to clarify, this was not the case when I queried directly against Solr (there I could resolve the entire tree). This, instead, just afflicts using `/catalog.json?q=` as the query endpoint.

jrgriffiniii:
Using `?q=` was breaking the query, removing this seems to return the children.

jrgriffiniii:
Indexing the parents into a different field (`tree_parent_ssim`) seems to be working as expected, however I'm not quite certain why this was necessary.

jrgriffiniii:
Queries should be optimized for nested Documents (ArcLight uses this), and I am working with Solr Document IDs as foreign keys - this approach is probably not appropriate (or just optimal) and should be changed.

[...]

PR #208, opened 2020-03-31
Updating the CircleCI configuration to use SimpleCov

jrgriffiniii:
This approach replicates what is found in https://github.com/pulibrary/figgy

PR #210, opened 2020-03-31
Reimplements the Collection Context interface using Vue

jrgriffiniii:
Advances #170


jrgriffiniii:
This should probably be rebased against #208 in order to ensure that the test coverage is proper for this.

jrgriffiniii:
Can any of this work be repurposed for the Table of Contents work? Perhaps the caching aspects of this can be helpful in advancing the implementation of the new design.

Otherwise, this should just be closed with statements regarding the what was deprecated.

jrgriffiniii:
I'm not certain where the work stands for the collection-caching, but the UI development work is now deprecated. I would please request that the branch be saved in order to port the Solr Document caching.

Issue #160, opened 2020-03-18
Ensure that all parent containers with dsc[2] information are indexed for Aeon requests


jrgriffiniii:
The ones with dsc2 description aren't getting indexed.

This is essential to the completion of #159

jrgriffiniii:
This blocked by indexing of the sub-containers within the dsc[2], previously this was just indexed as "Folder 2". This needs to have a link between the PULFALight staging and environment and PULFA 2.0.

jrgriffiniii:
This is a "normal" dsc, for contrast:
```
<dsc type="othertype" othertype="physicalholdings">
         <c level="otherlevel" otherlevel="physicalitem" id="C0040_i1">
            <did>
               <container type="box">1</container>
               <unitid type="barcode">32101040795617</unitid>
               <physloc type="code">rcpxm</physloc>
···············
            </did>
         </c>
      </dsc>
```

jrgriffiniii:
I'm going to resume working with indexing improvements in order to improve the performance for the new "Collection Context" interface work undertaken for https://github.com/pulibrary/pulfalight/issues/170.

jrgriffiniii:
Reviewing the work committed in https://github.com/pulibrary/pulfalight/pull/212, `dsc[2]` elements were actually getting indexed, and I misunderstood the original scope of this issue.

Currently, `components` are indexed within the top-level `collection` Solr Documents. This should document this and clearly outline how to access collection components.

I do not believe that `components` have child `components` within them indexed. That might need to be addressed for this as well.

jrgriffiniii:
https://github.com/pulibrary/pulfalight/commit/7064c7b95d464b9334cec6eb30c28e38b12a0577 seems to be the only proper approach which I could find for recursively processing child components of component elements (`<c>`).

tpendragon:
Recursively nested components are definitely indexed right now. There's a clipboard system that keeps track of parent nodes, and the `//` selector makes sure the selection is arbitrarily deep.

jrgriffiniii:
I was indexing components within the `components` field of the Solr Document with this approach, as I could only find this field populated for top-level collections.

jrgriffiniii:
More generally, I am honestly uncertain if this is even necessary to implement. I might close this issue in favor of another issue within https://github.com/pulibrary/pulfalight/issues/159

jrgriffiniii:
[...]

PR #218, opened 2020-04-01
Ensures that nested child components are nested within component Solr Documents

jrgriffiniii:
Ensures that component elements (<c>) with nested children have child components inserted into the "component" field of the Solr Document at the component level (rather than just the collection-level). Advances #160

### DIFF
--- a/lib/pulfalight/traject/ead2_config.rb
+++ b/lib/pulfalight/traject/ead2_config.rb
@@ -276,6 +276,16 @@ compose "components", ->(record, accumulator, _context) { accumulator.concat rec
                      record.attribute("id")&.value&.strip&.gsub(".", "-")
                    end
   end
+  to_field "components" do |record, accumulator, context|
+    child_components = record.xpath("./*[is_component(.)]", NokogiriXpathExtensions.new)
+    child_components.each do |child_component|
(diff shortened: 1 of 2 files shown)


tpendragon:
I sort of wonder if they weren't nested deeply on purpose for some reason? It seems like it was probably a choice to put them all at the parent level. Maybe someone knows?

jrgriffiniii:
The motivation which I had for this was to advance work for forming Aeon API requests within a component. My understanding is that requests will require this contextual information, as users might need to select either a component which is a direct child of a collection, or a descendent ("grandchild") for a component within a component.

My intention was that this information would be combined with the work done to index dsc[2] information in `physical_holdings`. The collection would have all dsc[2] indexed in `physical_holdings` for the document, but I doubt that we wanted to permit users to request an entire collection. The more realistic expectation I had was that users would be able to select a child component/descendent component for request, and that the `physical_holdings` field dat a would be combined with this in order to form the actual API.

Please let me know if this seems inappropriate, and also be aware that I'm not quite certain what is required for the Aeon request. I welcome any corrections to this (also, guidance from @sdellis is very welcome if you are available to comment).

jrgriffiniii:
Deprecated by #249

PR #219, opened 2020-04-02
Indexes dsc[2] elements into the collection-level documents in the "physical_holdings"

jrgriffiniii:
Advances #160, but does not address cases where dsc[2] information is indexed for individual components (should this include that?)

This is blocked by #218


lib/pulfalight/traject/ead2_config.rb
```diff
@@ -252,6 +253,218 @@
...
+  to_field "physical_location_ssm" do |_record, accumulator, context|
+    location_codes = context.output_hash.fetch("physical_location_code_ssm", [])
+    location_code = location_codes.first
+    next unless location_code
+
+    location = Pulfalight::PhysicalLocationResolver.resolve(location_code)
+    accumulator << location
+  end
+
+  # This is for indexing components
```
jrgriffiniii, line 295: I am still struggling with recursion here within the scope of Traject configuration. I have attempted one approach in which the I invoke the following:

```ruby
to_field "components" do |record, accumulator, context|
    child_components = record.xpath("./*[is_component(.)]", NokogiriXpathExtensions.new)
    child_components.each do |child_component|
      previous_parent = context.clipboard[:parent]
      context.clipboard[:parent] = record
      output = map_record(child_component)
      accumulator << output
      context.clipboard[:parent] = previous_parent
    end
  end
```

...as this approach was proposed in #218. However, I find that, when `map_record` in invoked within the scope of a `compose` block, the indexing steps *outside* of this block can no longer be run.

Perhaps through metaprogramming I can dynamically define indexing steps for fields attempt this approach for recursion (I avoided attempting this given the time consumed by difficulties in recursion I already experienced).

jrgriffiniii:
This currently has `physical_holdings` indexing the information as a child document:

image

Within the `components`, there are deeply nested documents for each of the components which link to the dsc[2] `components` containing the physical holding information:

image

With this, one could simply off a Solr query which retrieves only the `physical_holding` component child document, rather than its entire parent collection:

image

jrgriffiniii:
This needs to be improved from two standpoints:

- Containers without holding information need to be indexed also as child documents
- Components should have better information indicating whether or not it is a parent node or a leaf node in the hierarchy

jrgriffiniii:
The indexing for this is functional, but the `SolrDocument` does not permit me to retrieve these as desired. Currently, the adjusted query for the `fl` parameter is not being propagated to https://github.com/projectblacklight/blacklight/blob/v7.2.0/lib/blacklight/solr/repository.rb#L62.

jrgriffiniii:
Much of this is going to need to be adjusted in response to https://github.com/pulibrary/pulfalight/pull/249

Issue #183, opened 2020-03-19
Solr may not really be working in ci


hackmastera:
When I add logic to clean solr before each test all tests fail with a 404 from solr. I think this indicates that our ci setup is not complete.

jrgriffiniii:
I can confirm that this was the case for me at some point, I will try to reproduce this.

jrgriffiniii:
This does seem to still be a problem, as https://github.com/pulibrary/pulfalight/blob/master/.circleci/config.yml#L104 uses Solr for the CircleCI configuration, but does not seem to create a configset or core during test builds.

jrgriffiniii:
This was resolved with https://github.com/pulibrary/pulfalight/commit/49676f679202f315a5ad63a815f7bd3fbb57aa4d

Issue #100, opened 2020-01-22
Use the Blacklight range limit plugin for parsing and visualizing date ranges


jrgriffiniii:
While looking to facet upon date ranges or search for specific date ranges, UX design should be undertaken to assess and identify methods of indicating to users that a substantial number of items might be undated within collections

jrgriffiniii:
What should the warning for this look like? Wireframing will be needed in order to advance this.

jrgriffiniii:
A dialogue modal would be very disruptive - this should be in the location in the facet as a static message (perhaps using a Bootstrap alert color).

sdellis:
We could avoid any warning with a better representation of the data in the facet. The Date Range facet in  PulfaLight should be consistent with the Date Range facet in OrangeLight, which gives a histogram of date ranges in the results and a link to "undated" material that can't be represented. See screen shot (although I don't know why it spans beyond the current year). 

![Screen Shot 2020-03-30 at 12.02.31 PM.png](https://images.zenhubusercontent.com/5cba15f2e89c9c3b0c58e0d1/ae1140b9-d3c0-4126-8dc9-a555d954927d)

jrgriffiniii:
Invalid date years should not be rendered in the slider.

jrgriffiniii:
There are three possible cases which cannot be parsed into datestamps:

- Dates not examined (there may be dates but these were not found by those producing the EADs)
- Dates not found
- Dates which exist, but which cannot be parsed (e. g. "Easter Morning")

Facets will not directly map to the text which is rendered in the field for the record in all cases.

jrgriffiniii:
I can advance this by first installing the range limit plugin.

jrgriffiniii:
There are some problems with integrating this, as while it was apparently already configured on one of the facet fields, both the JavaScript is breaking, primarily due to the `blacklight_range_limit/_range_limit_panel.html.erb` partial not being rendered properly.

jrgriffiniii:
Resolved by #220

PR #220, opened 2020-04-03
Fixes date range limited faceting and introduces functional tests

jrgriffiniii:
Advances #100 and implements functional tests for the date range limited faceting using Capybara and Selenium. However, the Selenium tests are not currently passing given that the viewport cannot seem to have its width increased.


jrgriffiniii:
This doesn't seem to fail when Chrome is run without the `headless` option.

jrgriffiniii:
This is still breaking due to there needing to be an authenticated user creating the configset (please see https://lucene.apache.org/solr/guide/8_4/configsets-api.html#configsets-upload)

jrgriffiniii:
This is currently now passing with the usage of the Docker image (maintained at https://github.com/jrgriffiniii/pulfalight-solr-docker). I can definitely transfer ownership of this to `pulibrary` once this approach is found to be acceptable.

jrgriffiniii:
The Solr configuration changes introduced in this also should resolve https://github.com/pulibrary/pulfalight/issues/183

PR #222, opened 2020-04-03
Fix breadcrumbs

### LINKED ISSUE #217: Breadcrumb links are broken
The breadcrumbs are assuming the collection ID is in the link, which we've removed.

See https://pulfalight-staging.princeton.edu/catalog/C0396_c0335 for an example.

tpendragon:
Closes #217

### DIFF
--- a/app/models/arclight/parent.rb
+++ b/app/models/arclight/parent.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Arclight
+  ##
+  # Logic containing information about Solr_Ead "Parent"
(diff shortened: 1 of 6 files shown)


app/models/arclight/parent.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Arclight
```
jrgriffiniii, line 3: I have no objections to this, but I recall that `DigitalObject` was placed in `app/overrides` (please see https://github.com/pulibrary/pulfalight/blob/master/app/overrides/arclight/digital_object_override.rb). I would just please prefer that one approach be used consistently.

tpendragon: Done.

PR #229, opened 2020-04-07
Implementing the RobotsGeneratorService and adding a task for generating the robots.txt

jrgriffiniii:
This duplicates the work introduced by @eliotjordan into https://github.com/pulibrary/pulmap with https://github.com/pulibrary/pulmap/pull/553


public/robots.txt
```diff
@@ -1 +1,18 @@
-# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+User-agent: *
```
tpendragon, line 1: Shouldn't this be .gitignored, since it's generated on demand at deploy time?

PR #242, opened 2020-04-14
Implements JSON serialization for EAD collections

### LINKED ISSUE #241: Explore supporting EAD collection JSON serialization for the Table of Contents Vue components
This will support the JSON blobs integrated into the Table of Contents interface, as specified in https://github.com/pulibrary/pulfalight/issues/240#issuecomment-613117708

jrgriffiniii:
Resolves #241, and I believe that this should advance progressing towards developing the JSON objects which can be consumed by Vue components.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,6 +27,9 @@ Metrics/MethodLength:
     - 'lib/tasks/pulfalight.rake'
     - 'app/models/solr_document.rb'
     - 'app/controllers/catalog_controller.rb'
+Naming/PredicateName:
+  Exclude:
+    - 'app/models/solr_document.rb'
 RSpec/ExampleLength:
   Exclude:
     - 'spec/features/traject/ead2_indexing_spec.rb'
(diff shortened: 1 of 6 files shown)


jrgriffiniii:
While reviewing this with @khanhvu98kk and @sdellis, it was determined that attempting to provide objects (rather than strings) for child container nodes should also be supported in this (or another related PR).

jrgriffiniii:
Only the title and the ID is necessary in order to render each individual component. I will work to remove the unneeded attributes.

The nested hierarchy is actually what is valuable here, and the structure alone might have performance impacts for larger EADs.

jrgriffiniii:
This now also ensures that JSON-serialized Documents are retrieved using the `/catalog/:id/raw` route, as well as ensuring that the child components of a collection are serialized.

jrgriffiniii:
This is blocked by #249 as well as #248

jrgriffiniii:
After review from @sdellis, it has been determined that the fields for this need to be restricted to the following:

 - id
 - unittitle
 - has_digital_content (boolean)
 - components (array)

PR #243, opened 2020-04-14
Implements a minimal HTML view for EAD collections and components

### LINKED ISSUE #239: Explore implementing a shallow EAD component View using unstyled server-side HTML
This would be an alternative to implementing a JSON-API for data to be consumed by Vue components in the Table of Contents interface for navigating collections.

jrgriffiniii:
Resolves #239 and provides a recursive rendering of components which are child nodes of collections. This is currently a draft given that I have some proposed changes in this which I would like to push back to #242


jrgriffiniii:
This work might still be needed in order to support the breadcrumb generation.

jrgriffiniii:
This needs to be adjusted to provide breadcrumbs within the HTML snippet.

PR #248, opened 2020-04-14
Restructuring the DigitalObject override to use the Pulfalight namespace

### LINKED ISSUE #245: Restructure how app/overrides is used to override the DigitalObject Class
Also, we are currently using the `ArcLight` namespace, should we be doing that?

jrgriffiniii:
Resolves #245

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -18,12 +18,7 @@ module Pulfalight
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
-    # load overrides
-    config.to_prepare do
-      Dir.glob(Rails.root.join("app", "**", "*_override*.rb")) do |c|
-        Rails.configuration.cache_classes ? require(c) : load(c)
-      end
-    end
+    config.autoload_paths += Dir[Rails.root.join("lib", "pulfalight", "*.rb")]
 
     config.robots = OpenStruct.new(config_for(:robots))
   end
--- a/lib/pulfalight/digital_object.rb
+++ b/lib/pulfalight/digital_object.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # Add a role attribute to this class
-class Arclight::DigitalObject
+class Pulfalight::DigitalObject < Arclight::DigitalObject
   attr_reader :role
   def initialize(label:, href:, role:)
     @label = label.present? ? label : href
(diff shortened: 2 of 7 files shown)


lib/pulfalight/digital_object.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # Add a role attribute to this class
-class Arclight::DigitalObject
+class Pulfalight::DigitalObject < Arclight::DigitalObject
```
tpendragon, line 4: Why move this to lib, out of curiosity? Now we have models in two places. Does this work in like, `app/models/pulfalight/digital_object.rb`?

config/application.rb
```diff
@@ -18,12 +18,7 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
-    # load overrides
-    config.to_prepare do
```
tpendragon, line 22: Isn't there another override?

hackmastera:
@jrgriffiniii is it possible that spring was on and causing problems? Could the fix be as simple as turning off spring?

lib/pulfalight/digital_object.rb
jrgriffiniii: I was following the pattern in https://github.com/projectblacklight/arclight/blob/master/lib/arclight/digital_object.rb

config/application.rb
jrgriffiniii: I am sorry, I can adjust this.

jrgriffiniii:
@hackmastera I will try using a fresh clone and restarting spring.

jrgriffiniii:
@hackmastera I apologize, but stopping spring entirely still raised the error. The document against which I'm testing is `mudd/publicpolicy/MC104.EAD.xml`, and this error remains:

```
ArgumentError: unknown keyword: role
bundler/gems/arclight-6020c4ef4f65/lib/arclight/digital_object.rb:8:in `initialize'
```

I could try keeping this in place and just fixing the line on https://github.com/pulibrary/pulfalight/blob/master/lib/pulfalight/traject/ead2_config.rb#L13, but I just wanted to be certain that it might not be an issue elsewhere.

config/application.rb
hackmastera: There weren't any others.

hackmastera:
Oh, I see. The override load wasn't having an affect on the lib directory.

I think it would be safe to just fix it in that file, since anything in lib will always need an explicit require anyway.

jrgriffiniii:
Deprecated by #250

Issue #246, opened 2020-04-14
Support the serialization of containers as first-order objects


jrgriffiniii:
Currently the Traject configuration just indexes these as strings - they should probably also be indexed as nested Documents (in a manner similar to what is undertaken for the `components` field).

Blocks #242

jrgriffiniii:
This is going to be blocked by https://github.com/pulibrary/pulfalight/pull/249

jrgriffiniii:
This should probably not be the approach taken, as the JSON serialized need not be a direct mapping from the nested Solr Document tree.

PR #249, opened 2020-04-15
Ensures that Traject indexes nested component trees for collections with deeply-nested components

### LINKED ISSUE #247: Traject indexing fails for `components` in Collections with more than 1 level of descendent nodes
I recall encountering this when I was developing https://github.com/pulibrary/pulfalight/pull/236/files, and I can look to port the solution which I had introduced there.

jrgriffiniii:
Resolves #247

### DIFF
--- a/lib/pulfalight/traject/ead2_config.rb
+++ b/lib/pulfalight/traject/ead2_config.rb
@@ -286,8 +286,10 @@ compose "components", ->(record, accumulator, _context) { accumulator.concat rec
 
   to_field "ead_ssi" do |_record, accumulator, context|
     parent = context.clipboard[:parent]
+    next unless parent
+
     ead_ids = parent.output_hash["ead_ssi"]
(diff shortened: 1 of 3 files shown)


jrgriffiniii:
In order to advance #246, I have started refactoring the Traject configuration in https://github.com/pulibrary/pulfalight/compare/issues-247-jrgriffiniii-index-components-depth-refactor

jrgriffiniii:
This needs to take into account that queries for documents with nested components might not actually be possible without sending multiple queries against Solr; it might be best, instead, to index the entire tree into a string-serialized JSON of the component tree.

jrgriffiniii:
I'm testing this right now, and I'm actually finding that deeply-nested Solr Documents do get returned: https://gist.github.com/jrgriffiniii/150ff26d0826609d9e8468a1ad2133a6

PR #255, opened 2020-04-21
Refactors the Traject configuration for component indexing

jrgriffiniii:
This reorganizes the code base with the hope of simplifying adjustments for the indexing configuration.


jrgriffiniii:
After attempting various approaches to improve the performance of this, attempting to provide recursion for this on any level seems to just be slow. It might be necessary to improve this with the implementation of an asynchronous job for indexing the components alone.

PR #257, opened 2020-04-22
Implements indexing for <container> elements as nested Solr Documents

### LINKED ISSUE #246: Support the serialization of containers as first-order objects

jrgriffiniii:
Resolves #246

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,6 +145,9 @@ jobs:
       - store_artifacts:
           path: /tmp/test-results
           destination: test-results
+      - store_artifacts:
+          path: ~/pulfalight/coverage
+          destination: coverage
 
   coverage_report:
     working_directory: ~/pulfalight
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,6 +16,7 @@ AllCops:
 Metrics/AbcSize:
   Exclude:
     - 'lib/pulfalight/traject/ead2_component_config.rb'
+    - 'lib/pulfalight/traject/ead2_container_config.rb'
 Metrics/BlockLength:
   Exclude:
     - 'spec/**/*'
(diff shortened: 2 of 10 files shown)


jrgriffiniii:
The `rspec` test suite is now consuming 12 minutes on CircleCI - perhaps one of the test fixtures is the root cause of this problem.

jrgriffiniii:
The coverage analysis is not accurate given that the `rspec` test suite is exceeding 10 minutes in length:

```
Too long with no output (exceeded 10m0s): context deadline exceeded 
```
(please see https://app.circleci.com/pipelines/github/pulibrary/pulfalight/432/workflows/1579d98a-9206-4ff5-8b52-4171c2e3106e/jobs/661)

For various reasons, the test suite needs to be simplified in order to resolve this.

jrgriffiniii:
It seems as if running Traject twice, even on the same EAD file, triggers a massive delay in the tests being completed.

jrgriffiniii:
Performance for this is being significantly increased merely by the introduction of `context` blocks. Testing this locally, I used one case with and without an unnecessary `context` block:

Without the `context`:
`Finished in 1 minute 47.55 seconds (files took 5.2 seconds to load)`
With the new `context`:
`Finished in 2 minutes 9.6 seconds (files took 6.42 seconds to load)`

Given that this is testing the Traject configuration, I can look to try and just reduce the number of these `context` blocks, and reduce the complexity of the EAD fixtures.

jrgriffiniii:
The recursive approach introduced in https://github.com/pulibrary/pulfalight/pull/255 is what is triggering the delays. This will need to be adjusted.

jrgriffiniii:
@jrgriffiniii needs to rebase this first, but will request a later review.

jrgriffiniii:
This should not be merged until the breadcrumbs are fixed. If this blocks breadcrumbs from being rendered, then this should be closed.

jrgriffiniii:
Unless this can be reconciled with ArchivesSpace EADs (which already have container IDs), this is no longer necessary.

PR #264, opened 2020-05-01
Indexes and renders collection-level notes on collection pages and for component pages

### LINKED ISSUE #224: Collection-level notes
Collection-level notes (notes currently included in the Overview and Access tabs on pulfalight) should include all elements/notes that appear in archdesc from the opening did element to the first dsc combined tag. (See all tabs associated with finding aids on the current site: Summary, Description, Collection History, etc.) 

The "Rules or Conventions" note should not appear.

As indicated in #119, most of these notes should filter down to child component records unless those notes are supplanted at lower levels.

jrgriffiniii:
Resolves #224

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -4,6 +4,21 @@ class SolrDocument
   include Arclight::SolrDocument
   include ActiveModel::Serialization
 
+  # self.unique_key = 'id'
+
+  # Email uses the semantic field mappings below to generate the body of an email.
+  SolrDocument.use_extension(Blacklight::Document::Email)
+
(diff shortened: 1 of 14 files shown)


app/models/solr_document.rb
```diff
@@ -40,6 +55,23 @@ def collection?
     level == self.class::COLLECTION_LEVEL
   end
 
+  def component?
+    key?("component_level_isim")
+  end
+
+  def html_presenter_class
```
tpendragon, line 62: It looks like the implementation of both of these are the same - why do we need two presenters?

jrgriffiniii:
Deprecated by https://github.com/pulibrary/pulfalight/pull/279

Issue #79, opened 2019-11-11
Users should be able to search by collection identifiers


jrgriffiniii:
This feature was long desired on the PULFA 2.0 application, but due to various constraints, this was never implemented. This might already be supported with the current approach to indexing and Blacklight configuration, but including this in cognitive walkthroughs for user testing might still be desirable.

jrgriffiniii:
After discussing this with @faithc, it was confirmed that the immediate need is to ensure that users are able to search by component identifiers (`<c id="...">`). Container-level searching requires a larger discussion of the data model being used in the EAD.

jrgriffiniii:
This needs to have acceptance criteria in order to be properly closed.

jrgriffiniii:
This can be achieved, where component IDs are already structured in the data with the collection ID prefixed.

jrgriffiniii:
**Acceptance criteria:** When users search for "AC044_c0001", the targeted component record is retrieved in the search results.

Please be aware that the padding with 0s is not consistent between collections (users cannot try to guess these).

khanhvu98kk:
Resolved by pulibrary/pul_solr#210

tpendragon:
Closed as duplicate of #97

PR #266, opened 2020-05-01
Ensuring that users are redirected to the show page when searching for a valid collection ID

jrgriffiniii:
Ensuring that, when users search for a collection ID, that they are redirected to the collection show page (should the collection exist). Advances #79

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -38,6 +38,20 @@ class CatalogController < ApplicationController
     end
   end
 
+  def index
+    query_param = params.fetch(:q)
+    return super unless query_param =~ /^[A-Z]{1,2}\d{3,4}$/
+
+    # Try and take the user directly to the show page
+    begin
+      _deprecated_response, @document = search_service.fetch(query_param)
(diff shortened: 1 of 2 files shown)


tpendragon:
@apferrar @faithc Can you confirm that when I search for WC063 I should end up not at search results, but instead at the collection page? I'm not seeing that use case reflected in #79 and it feels a little inconsistent and potentially surprising to me, but I'm not a common user of finding aids.

faithc:
Hi @tpendragon - thanks for checking in about this. Our initial request (#79) was that the collection just show up as the first result on the search results page per the behavior of the current site; however, I do think that taking one directly to the collection makes more sense to me since that is what the user is looking to do. Taking them to a search results page seems an unnecessary step since from there they just click on the first result. Curious what @apferrar thinks.

apferrar:
@tpendragon @faithc good question! I can see pros and cons but let's go for it and see how change works!

jrgriffiniii:
Deprecated by the rebase for https://github.com/pulibrary/pulfalight/pull/280

PR #280, opened 2020-06-30
Ensuring that users are redirected to the show page when searching for a valid collection ID

jrgriffiniii:
Ensuring that, when users search for a collection ID, that they are redirected to the collection show page (should the collection exist). Advances #79

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -38,6 +38,20 @@ class CatalogController < ApplicationController
     end
   end
 
+  def index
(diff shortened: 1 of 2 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -38,6 +38,20 @@ def show
     end
   end
 
+  def index
+    query_param = params.fetch(:q)
```
eliotjordan, line 42: issue: This raises an error when the query param is empty.
Replacing with `params[:q]` fixes the issue.

PR #293, opened 2020-07-09
Updating the Traject config. to extract barcodes from <container> elements and indexing them into the barcodes_ssim field

### LINKED ISSUE #91: Resolve location information in EADs for requests
EADs for the Manuscripts Division material have <dsc> elements along with <dsc2> tracking physical item information, also containing physical location information (absolute identifier or a box number, there may also be text indicating where, in the vault, material might be located). Please be aware that this does not affect the Mudd Library.

When user request containers, there needs to be location information which is provided along with the Aeon requests in order to ensure that requests can be fulfilled by staff.

jrgriffiniii:
Resolves #91

### DIFF
--- a/lib/pulfalight/traject/ead2_component_config.rb
+++ b/lib/pulfalight/traject/ead2_component_config.rb
@@ -330,6 +330,18 @@ to_field "collection_notes_ssm" do |_record, accumulator, context|
   accumulator.concat(parent.output_hash["collection_notes_ssm"])
 end
 
+to_field "barcodes_ssim" do |record, accumulator|
+  record.xpath("./did/container[@label]").each do |node|
+    label = node.attr("label")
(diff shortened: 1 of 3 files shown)


lib/pulfalight/traject/ead2_component_config.rb
```diff
@@ -325,6 +325,18 @@
...
 
+to_field "barcodes_ssim" do |record, accumulator|
+  record.xpath("./did/container[@label]").each do |node|
+    label = node.attr("label")
+    barcode_match = label.match(/\[(\d+?)\]/)
+
+    if barcode_match
+      barcode = barcode_match[1]
+      accumulator << barcode
+    end
```
eliotjordan, line 336: suggestion: collapse this logic into a simple one liner
```
accumulator << barcode_match[1] if barcode_match
```

PR #297, opened 2020-07-10
Updating the Traject config. to extract barcodes from <container> elements and indexing them into the barcodes_ssim field

### LINKED ISSUE #91: Resolve location information in EADs for requests

jrgriffiniii:
Resolves #91 and deprecates #293 with a rebase.


jrgriffiniii:
https://github.com/pulibrary/pulfalight/pull/293 did not need to be deprecated

PR #299, opened 2020-07-10
Update pulfalight:seed task to run indexing inline and commit immediately

### LINKED ISSUE #292: Add task to index fixture documents into development instance
- Would be useful to have a quick way of indexing fixture docs into a development solr instance with a single rake task command.
- Should index index inline so we don't have to launch sidekiq if the index time isn't too long.

eliotjordan:
Closes #292

### DIFF
--- a/lib/tasks/pulfalight.rake
+++ b/lib/tasks/pulfalight.rake
@@ -75,12 +75,10 @@ namespace :pulfalight do
   end
 
   desc "Seed fixture data to Solr"
-  task :seed do
+  task seed: :environment do
     puts "Seeding index with data from spec/fixtures/ead..."
-
-    Dir.glob("spec/fixtures/ead/**/*.xml").each do |file|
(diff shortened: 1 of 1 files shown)


lib/tasks/pulfalight.rake
```diff
@@ -75,12 +75,10 @@ namespace :pulfalight do
...
   desc "Seed fixture data to Solr"
-  task :seed do
+  task seed: :environment do
     puts "Seeding index with data from spec/fixtures/ead..."
-
-    Dir.glob("spec/fixtures/ead/**/*.xml").each do |file|
-      index_file(relative_path: file, root_path: Rails.root)
-    end
+    index_directory(name: "spec/fixtures/ead/", root_path: Rails.root, enqueue: false)
+    blacklight_connection.commit
```
jrgriffiniii, line 81: 👍

PR #301, opened 2020-07-10
Ensuring that the collection-level extent is not just the first value of extent_ssm and rendering these values as additional badges on the search results view

### LINKED ISSUE #127: Collection level extents not displaying consistently.
Here's an example: https://pulfalight-staging.princeton.edu/?group=true&q=american+civil+liberties+union&search_field=all_fields&view=compact  group by collection, you can see that ACLU 2, Lillenthal, and Doar are all different. Faith checked the data and it seemed to be consistent so she thinks it has to do with the site.

jrgriffiniii:
Resolves #127, please see the following screenshot for a preview:

image

### DIFF
--- a/app/views/catalog/_arclight_index_default.html.erb
+++ b/app/views/catalog/_arclight_index_default.html.erb
@@ -1,0 +1,48 @@
+<div class="al-search-result-index-article row d-flex align-items-start">
+  <% counter = document_counter_with_offset(document_counter) %>
+  <div class="col-auto">
+    <%= render partial: 'arclight_document_header_icon', locals: { document: document }  %>
+  </div>
+
+  <div class="col col-no-left-padding d-flex flex-wrap">
+    <div class="documentHeader my-w-75 w-md-100 order-0" data-document-id="<%= document.id %>">
+      <h3 class="index_title document-title-heading">
+        <%= link_to_document document, document_show_link_field(document), counter: counter %>
+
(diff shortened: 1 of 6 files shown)


app/views/catalog/_arclight_index_default.html.erb
```diff
@@ -0,0 +1,48 @@
...
+  <div class="col-auto">
+    <%= render partial: 'arclight_document_header_icon', locals: { document: document }  %>
+  </div>
+
+  <div class="col col-no-left-padding d-flex flex-wrap">
+    <div class="documentHeader my-w-75 w-md-100 order-0" data-document-id="<%= document.id %>">
+      <h3 class="index_title document-title-heading">
+        <%= link_to_document document, document_show_link_field(document), counter: counter %>
+
+        <% document.extents.each do |extent| %>
```
eliotjordan, line 12: suggestion (possibly blocking): I wonder if we should propose this change upstream to avoid having to override the entire index_default partial. Do you think that there would be support for displaying multiple extents? The field is `ssm` after all.

eliotjordan:
@jrgriffiniii I created a ticket in Arclight for potentially pushing this upstream.

https://github.com/projectblacklight/arclight/issues/997

eliotjordan:
Ill merge this for now and then we can update if/when the upstream changes happen.

PR #312, opened 2020-07-17
Index fixtures as Solr documents

### LINKED ISSUE #310: Load fixture documents when starting solr
This is a more conventional method of setting up a blacklight test environment.

Advantages:

- Because the current specs are mocking Solr responses in some cases, there are some divergences with what Solr actually returns and what the spec is testing for. Preloading all the specs into Solr and using those documents resolves this issue. See: https://github.com/pulibrary/pulfalight/blob/master/spec/requests/catalog_request_spec.rb#L6-L41

- This could possibly reduce test time.

Partially addresses some issues raised in #307

eliotjordan:
Closes #310 

- Creates a wrapper around lando `bin/solr` to index fixtures into testing solr.
- Loads fixtures into test and dev solr_wrapper instances.
- Updates specs to use solr documents rather than mocked Solr responses.
- Updates bixby/rubocop and appeases the officers.
- Lowers test time from 2 minutes 12 seconds to 44 seconds.


eliotjordan:
**Note for reviewers**. If you use lando, you'll have to run `lando stop` and then `bundle exec solr start` to get the fixtures loaded correctly.

app/overrides/arclight/digital_object_override.rb
```diff
@@ -4,7 +4,7 @@
 class Arclight::DigitalObject
   attr_reader :role
   def initialize(label:, href:, role:)
-    @label = label.present? ? label : href
+    @label = label.presence || href
```
jrgriffiniii, line 7: 👍

PR #325, opened 2020-07-24
Introduces support for Aeon form API integration

### LINKED ISSUE #315: Integrate support for the Aeon form API
This integration uses https://lib-aeon.princeton.edu/aeon/Aeon.dll in order to submit requests against Aeon, and will borrow heavily from the [PULFA 2.0 Aeon support](https://github.com/PrincetonUniversityLibrary/pulfa/blob/production/webapp/xsl/component2html.xsl#L292)

jrgriffiniii:
Resolves #315


jrgriffiniii:
Please note that while the "Request" button is unstyled and enabled for all Collections and Components in this: 
image

I was hesitant to introduce any UI development work which might block this with #327

app/values/pulfalight/location_code.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module Pulfalight
+  class LocationCode
```
jrgriffiniii, line 4: Please see https://github.com/PrincetonUniversityLibrary/pulfa/blob/production/webapp/xsl/lib.xsl#L52 for where these values were originally derived.

app/views/arclight/requests/_aeon_external_request_endpoint.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= form_tag @aeon_external_request.url, method: :post, class: 'al-request-form' do |f| %>
```
jrgriffiniii, line 1: This is going to need to be deprecated once the Request Cart Vue Component is integrated.

app/helpers/pulfalight_helper.rb
```diff
@@ -23,4 +23,14 @@ def collection_notes_formatter(*_args)
   def component_notes_formatter(*_args)
     html_presenter.collection_notes
   end
+
+  # @return [Class]
+  def aeon_external_request_class
+    Pulfalight::Requests::AeonExternalRequest
```
jrgriffiniii, line 29: It might be worth enquiring about ensuring that the class used for Aeon requests can be configured somewhere upstream in ArcLight.

PR #358, opened 2020-09-02
allows keyboard-only TOC navigation

sdellis:


eliotjordan:
Close in favor of https://github.com/pulibrary/pulfalight/pull/373

Issue #364, opened 2020-09-04
Integrate the Universal Viewer into the Component Show View


jrgriffiniii:
This was not ported when the new design implementation for the Component Show View was merged.

jrgriffiniii:
There is also behavior which should be ported from PULFA 2.0, specifically in which digital objects dynamically update the state of the UV.

tpendragon:
Closed by #369, split out #389

PR #369, opened 2020-09-09
Display Universal Viewer

tpendragon:
Work towards #364 

This doesn't quite finish things up because it doesn't include the JS part which queries Figgy, but increasing the size of the layout seemed controversial enough that I figured I should put it up sooner rather than later. Attached is a screenshot of the new viewport on my macbook.

Screen Shot 2020-09-09 at 4 52 43 PM


app/views/catalog/_component_summary.html.erb
```diff
@@ -14,6 +14,8 @@
     </div>
 
   </div>
-
+  <% if document.digital_objects.present? %>
+    <%= render(partial: "arclight_viewer_default") %>
```
eliotjordan, line 18: There should be a follow up for dealing with un-styled digital object links. E.g. `catalog/MC221_c0118`

eliotjordan (approved):

PR #373, opened 2020-09-10
Replace jqtree widget with jstree and use turbolinks

### LINKED ISSUE #360: Use jstree widget for table of contents
The library we currently use (jqtree) is cannot be navigated easily with a keyboard and does not meet accessibility standards as a result.

### LINKED ISSUE #362: Clicking the Leaf node on the Table of Contents should reload the content section dynamically
Currently the behavior of the ToC is such that clicking a leaf node refreshes the entire page.

eliotjordan:
Replaces `jqtree.js` with `jstree.js` for better accessibility. Tree can now be navigated using the keyboard.

- Uses a vendored copy of jstree because using yarn appears to add a second copy of jquery.
- Getting jstree to play nice with the asset pipeline is tricky and required the creation of a custom css theme.

Also adds turbolinks for better user experience with TOC.

Closes #360 
Closes #362


app/assets/javascripts/toc.js
```diff
@@ -1,28 +1,88 @@
...
-      initialUrl,
-      function(data) {
-        element.tree({
-          data: data
-        });
-      }).done(function() {
-        let selectedNode = element.tree('getNodeById', selectedId);
-        element.tree('selectNode', selectedNode);
-    });
+class TocBuilder {
```
tpendragon, line 1: Suggestion: Should we change the extension of this file to `.es6`? Also, should this be in `app/javascript`? I don't know if we've gotten webpack running with this project yet, though. If not we don't have to worry about it.

app/assets/javascripts/toc.js
```diff
@@ -1,28 +1,88 @@
...
+  }
+}
+
+
+// Setup table of contents on initial page load
+$(document).ready(function() {
+  const toc = new TocBuilder('#toc')
+  toc.build()
+  // Set initial toc build flag
+  window.buildToc = true
```
tpendragon, line 65: Suggestion: Instead of just a flag maybe put the `toc` variable in the global space and check to see if it's defined?

app/assets/javascripts/toc.js
```diff
@@ -1,28 +1,88 @@
...
-          let node = event.node
-          let url = `/catalog/${node.id}`
-          window.location.href = url
+  setupEventHandlers() {
+    // Listen for click on the leaf node and follow link to component
+    this.element.on('activate_node.jstree', function (e, data) {
+      let location = `/catalog/${data.node.id}`
+      Turbolinks.visit(location)
+      // Scroll to top of page instead of jumping for better user experience
+      window.scrollTo({ top: 0, behavior: 'smooth' });
```
tpendragon, line 33: Question: Is there a way to make it scroll up after the page loads? I don't know if it'll be like this in prod, but I run into this issue where it scrolls up to old content and just looks like it's broken.

tpendragon (changes requested):

app/assets/javascripts/toc.js
eliotjordan: Using the `.es6` extension doesn't work. Compilation throws an error about babelHelpers that I can't resolve.

I can try webpack and see how that works.

app/assets/javascripts/toc.js
eliotjordan: Are you suggesting that we define buildToc outside of this function/class?

```
let buildToc = true
class TocBuilder {...
```

app/assets/javascripts/toc.js
eliotjordan: You mean it scrolls before the refreshed content has finished loading?

tpendragon: Yeah.

app/assets/javascripts/toc.js
tpendragon: No, I'm suggesting `window.toc = new TocBuilder('#toc')`

eliotjordan: We still need a flag here. Determining if a TocBuilder is defined is not enough. We want to rebuild the TOC when navigating turbolinks via history (forward /back) but NOT after clicking a link and triggering a turbolinks load.

app/assets/javascripts/toc.js
eliotjordan: Getting smooth scroll to play nice with turbolinks rendering is a challenge. I'm going to remove it and just default back to standard behavior for now. I think it looks fine without it. We can revisit at a later date if it becomes an issue.

tpendragon (approved):

PR #395, opened 2020-09-23
Make collection title a link to the top-level collection record

### LINKED ISSUE #393: Make collection title a link to the top-level collection record
Without breadcrumbs, it has become more difficult to navigate to the top-level collection. One solution for this is to create collection title anchor that links to the record.

eliotjordan:
Closes #393

### DIFF
--- a/app/views/catalog/_show_collection_title.html.erb
+++ b/app/views/catalog/_show_collection_title.html.erb
@@ -1,7 +1,9 @@
 
 <div class="collection title" aria-label="collection title">
   <%= content_tag :h1, class: "media" do %>
-    <span class="col"><%= document.collection.first %></span>
+    <%= content_tag :a, href: collection_link, data: { turbolinks: "false" } do %>
+      <span class="col"><%= document.collection.first %></span>
+    <% end %>
   <% end %>
 
   <div class="collection-id">
--- a/app/views/catalog/_show_collection_title.html.erb
+++ b/app/views/catalog/_show_collection_title.html.erb
@@ -1,7 +1,11 @@
 
 <div class="collection title" aria-label="collection title">
   <%= content_tag :h1, class: "media" do %>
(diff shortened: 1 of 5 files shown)


app/views/catalog/_show_collection_title.html.erb
```diff
@@ -1,7 +1,9 @@
 
 <div class="collection title" aria-label="collection title">
   <%= content_tag :h1, class: "media" do %>
-    <span class="col"><%= document.collection.first %></span>
+    <%= content_tag :a, href: collection_link, data: { turbolinks: "false" } do %>
```
tpendragon, line 4: Suggestion: Can we make this
```ruby
<%= link_to_if(document.ead.present?, solr_document_path(document.ead), data: { turbolinks: "false" }) do %>
<span class="col"><%= document.collection.first %></span>
<% end %>
```

That'll make it not add the anchor if there's no EAD property, and avoids needing the helper.

tpendragon (changes requested):

app/views/catalog/_show_collection_title.html.erb
eliotjordan: I forgot about `solr_document_path` for some reason. However, there is a long-standing issue with `link_to_if` where it does not render the block as you might expect: https://github.com/rails/rails/issues/14830#issuecomment-42476475

If the desire is to avoid adding a helper method, we can add the logic into the partial, which I'll try and push up as an amendment. 

```
  <% if document.ead.present? %>
    <%= link_to solr_document_path(document.ead), data: { turbolinks: "false" } do %>
      <span class="col"><%= document.collection.first %></span>
    <% end %>
  <% else %>
    <span class="col"><%= document.collection.first %></span>
  <% end %>
```

tpendragon: I'm personally for not having the helper method for this.

tpendragon (approved):

PR #400, opened 2020-10-02
Retrieve library hours from LibCal

### LINKED ISSUE #378: Make Homepage Hours pull from libcal

eliotjordan:
Closes #378 

Retrieves library hours from LibCal. Because of CORS restrictions that we can't fix (Springshare manages the servers), hours data is pulled in and transformed server side.

Also:

- Adds link to the more hours button
- Organizes javascript with loader classes

Screen Shot 2020-10-02 at 9 20 55 AM


app/javascript/pulfalight/toc_loader.es6
```diff
@@ -0,0 +1,46 @@
+import TocBuilder from "../pulfalight/toc.es6"
+
+export default class TocLoader {
+  run() {
+    this.setup_toc()
+    this.setup_before_visit_listenter()
```
tpendragon, line 6: Really small typo - should be `listener`

app/services/hours_builder.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  attr_reader :id
+  def initialize(id:)
+    @id = id.to_i
+  end
+
+  def build
+    # Return if id parameter is not an integer
+    return hours_hash(unvailable_message) if id.zero?
+    response = query_lib_cal
```
tpendragon, line 16: Suggestion: Extract to memo-ized instance method.

app/services/hours_builder.rb
```diff
@@ -0,0 +1,57 @@
...
+    # Return if LibCal can't be reached
+    return hours_hash(unvailable_message) unless response
+    value = parse_response(response)
```
tpendragon, line 19: Suggestion: Extract to memo-ized instance method.

app/services/hours_builder.rb
```diff
@@ -0,0 +1,57 @@
...
+  def self.build(id:)
+    new(id: id).build
+  end
+
+  attr_reader :id
+  def initialize(id:)
+    @id = id.to_i
+  end
+
+  def build
```
tpendragon, line 13: Suggestion: Extracting some of these methods and making `value` return `hours_hash(unavailable_message)` (or nil I guess) if the id is zero or there's no response will tighten up this logic considerably and make it look nicer and read quicker, I think.

tpendragon (changes requested):
Couple small things.

app/services/hours_builder.rb
eliotjordan: Refactored.

tpendragon (approved):

PR #418, opened 2020-10-07
Group search results by collection by default

### LINKED ISSUE #113: Revert "Group search results by collection as a default (rather than the "All results" view)"
The "all results" view is too busy and it looks better when automatically grouped by collection. There should still be the "All results" view as an option, just not the default.

eliotjordan:
Closes #113


tpendragon (changes requested):
Failing CI, but also it shouldn't group by collection if it's searching "in this collection"

eliotjordan:
@tpendragon Updated to not group when faceting by collection.

tpendragon (approved):

PR #427, opened 2020-10-13
Index from ArchivesSpace

tpendragon:
This is just a first pass and the the very beginning, but it should get some EADs in for us.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -66,6 +66,7 @@ group :development, :test do
   gem "database_cleaner"
 end
 
+gem "archivesspace-client", github: "pulibrary/archivesspace-client", branch: "fix_login"
 gem "arclight", git: "https://github.com/projectblacklight/arclight.git"
 gem "blacklight_range_limit", "~> 7.1"
 gem "bootstrap", ">= 4.3.1"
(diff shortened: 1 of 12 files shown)


Gemfile
```diff
@@ -66,6 +66,7 @@ group :development, :test do
   gem "database_cleaner"
 end
 
+gem "archivesspace-client", github: "pulibrary/archivesspace-client", branch: "fix_login"
```
jrgriffiniii, line 69: 👍 for the shifting the version requirements.

app/jobs/aspace_index_job.rb
```diff
@@ -0,0 +1,60 @@
...
+    repository = Blacklight.default_index
+    repository.connection
+  end
+
+  def logger
+    Rails.logger || Logger.new(STDOUT)
+  end
+
+  def aspace_client
+    @aspace_client ||= Aspace::Client.new
```
jrgriffiniii, line 42: This is not a criticism, but I would be curious as to whether or not a future feature might be to provide a `Configuration` Object for the `Aspace::Client` constructor. This is not needed here.

jrgriffiniii: I am sorry, please disregard this, I will delete these comments looking below.

jrgriffiniii (approved):
Thank you very much for this, I've created https://github.com/pulibrary/princeton_ansible/issues/1788 in response to the new environment variables used for ArchivesSpace authentication.

Issue #89, opened 2020-01-07
Implement the design for the Expanded Collection View defined by the Summer 2020 wireframes


jrgriffiniii:
This should enable users to be able to perform a CTRL+F browser-based search for a string for content within any given component. *This should be supported for all finding aids in the app.*

If a user is in a show page for a component, they should be able to navigate to the single-page view of the Finding Aid.

jrgriffiniii:
ArcLight does not come with this out of the box, hence, we are going to need to implement an entirely separate view for this.

jrgriffiniii:
This is termed as the "expanded view" in the design documents.

jrgriffiniii:
This also ensures that users can use in-browser search (CTRL+F) in order to search within the finding aids.

jrgriffiniii:
Please see https://xd.adobe.com/view/c3b49814-866d-449f-44e7-2479ff0cbb30-b585/screen/7366cbcf-3a86-41ae-86e6-432e3d1e35b6 for the design of this interface.

tpendragon:
Closed by #435

PR #435, opened 2020-10-23
Implements a new view for an expanded finding aid collection

jrgriffiniii:
Advances #89, but I would like to first please request some feedback before finalizing this.


jrgriffiniii:
There was a bug with this originally, and only a limited number of child documents were rendered. Also, the Table of Contents widget was not expanded:

image

jrgriffiniii:
This needs to be rebased.

eliotjordan (commented):
@jrgriffiniii A few thoughts / questions:

1. We will need an expanded view button on the TOC header bar to be compliant with the wireframes.
1. The component titles should be links that take the user to the component. Right?
1. I wonder about expanding the TOC. The wireframe has it closed and it just seems like unnecessary and duplicative information to me.
1. From the wireframes, it looks like we need more metadata for components at a certain level. Looks like "Scope note" in the drawings. I'm not sure if that's correct or just filler.

jrgriffiniii:
https://github.com/pulibrary/pulfalight/pull/435/commits/59ff113557183ebde1fd87dda4717177b41a78d7 introduces the metadata for components with `scope note` and `arrangement`, as well as an `expand` and `collapse` button:
image

image

jrgriffiniii:
https://github.com/pulibrary/pulfalight/pull/435/commits/9f766d2b47b2841b75bd4157b001ade966ca4bda ensures that component titles are links and improves the styling for the extent badges:

image

jrgriffiniii:
https://github.com/pulibrary/pulfalight/pull/435/commits/cbe210e9f31b57a09ca56e18c883eeb40534727f ensures that the Table of Contents component is collapsed for the expanded Collection View.

jrgriffiniii:
Rebasing this introduced a number of test failures which I'm now having difficulty reproducing locally. I suspect that this might be due to some updates applied to the EAD XML fixtures used during some the tests.

eliotjordan (changes requested):
@jrgriffiniii  I have one code comment so far and a serious concern about performance. After experimenting a little, it seems that the SolrDocumentTree is making a single solr query for each child document. This works OK in dev with fixtures, but caused a serious problem on staging when trying to build the tree for the full Toni Morrison collection. The way I solved this for the TOC was to make one Solr request for the entire collection document with all of it's nested children and then parse that in the Ruby.

spec/models/solr_document_tree_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  let(:values) do
+    indexer.map_record(record)
+  end
+  let(:solr_document) do
+    SolrDocument.new(values)
+  end
+
+  context "when the Solr Document contains one or more member components" do
+    let(:fixture_path) do
+      Rails.root.join("spec", "fixtures", "aspace", "generated", "publicpolicy", "MC152.EAD.xml")
```
eliotjordan, line 39: No need to index this record again. It's already a fixture. On line 34, you could just use `SolrDocument.find("MC152")`

jrgriffiniii:
https://pulfalight-staging.princeton.edu/catalog/MC120?expanded=true is still quite slow even with https://github.com/pulibrary/pulfalight/pull/435/commits/0b2440f1772e7d1c14178d5c747e17ceb409fe9d, but there were definitely improvements. My suspicion is that this might also just be traversing of the nested Solr Documents.

eliotjordan (approved):

PR #438, opened 2020-10-27
Adds Focus to Homepage Searchbox

### LINKED ISSUE #437: Set focus to search box on homepage
A user should be able to simply start typing their search terms when they land on the homepage.

sdellis:
fixes #437

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -27,4 +27,4 @@
 // For blacklight_range_limit built-in JS, if you don't want it you don't need
 // this:
 //= require 'blacklight_range_limit'
-//= requiere 'focus'
+//= require 'focus'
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/application.js
```diff
@@ -27,3 +27,4 @@
 // For blacklight_range_limit built-in JS, if you don't want it you don't need
 // this:
 //= require 'blacklight_range_limit'
+//= requiere 'focus'
```
tpendragon, line 30: Shouldn't this be `//= require`?

tpendragon (changes requested):

app/assets/javascripts/application.js
sdellis: Yes, absolutely. I'm actually surprised it still works. Is it necessary?

tpendragon (approved):

PR #439, opened 2020-10-27
Request Cart

tpendragon:
This copies over the work that @jrgriffiniii did and replaces some of the Ruby so it's tested and all in one place.

We should test the JS components too, but I haven't gotten there yet.


sdellis:
@tpendragon a few thoughts: 

- Can you provide a list of the scenarios you are testing for, along with the fixtures and component ids that represent them? I'm having trouble finding components to test with in the subset of data we load.
- One should not be able to add an entire series to the cart, but should instead be able to select from a pick-list of child components.
- The cart detail should not scale in width based on the viewport size. On a mobile device, it should take up the entire screen.
- Somehow the icons look different in the header CartToggle component and the CartDetail and AddToCart components. 
- The AddToCart button should be visibly disabled or replaced with "Item added to cart" text if the item has already been added to the cart.
- The "Located In" field only lists subcontainers, but it should also list the container. I'm also curious how the container gets passed to Aeon as I don't see it when I inspect the form that gets posted to Aeon. When requesting C1588_c4, Folder 2 is listed as both the container and subcontainer in Aeon.
- The title in the CartDetail component should be the full title. It seems to cut off the dates that would distinguish it from other items. (Looking at C1588_c4 for reference.)

tpendragon:
@sdellis Sorry, I've been using ASpace records. We haven't nailed down fixtures from ASpace yet.

tpendragon:
@sdellis I created https://github.com/pulibrary/pulfalight/issues/440, https://github.com/pulibrary/pulfalight/issues/441, and https://github.com/pulibrary/pulfalight/issues/442 as follow ups so we can handle those things in parallel and smaller chunks after this is merged. Seem okay?

sdellis:
@tpendragon I agree, it's good to move those into separate tickets. Commented on one after thinking more about it.

app/javascript/components/RequestCart.vue
```diff
@@ -0,0 +1,522 @@
...
+          <caption>
+
+    <input-button
+      v-on:button-clicked="toggleCartView($event)"
+      type="button"
+      variation="text"
+      class="denied-button"
+    >
+
+      <div class="lux-icon">
```
sdellis, line 20: This should use the same icon that is used in other RequestCart components. I noticed that there are both CartIcon and RequestCartIcon that have been added in this PR.

app/assets/stylesheets/modules/_request_cart.scss
```diff
@@ -0,0 +1,6 @@
+@mixin request-cart {
```
eliotjordan, line 1: question: This is an empty block; is it here for future expansion/styling?

app/javascript/components/CartViewToggle.vue
```diff
@@ -0,0 +1,63 @@
...
+        return this.cart.items
+      }
+    },
+    ...mapState({
+      cart: state => store.state.cart,
+    })
+  },
+  methods: {
+    toggleCartView(event) {
+      console.log(event)
```
eliotjordan, line 38: issue: Should remove.

app/javascript/components/RequestCart.vue
```diff
@@ -0,0 +1,522 @@
...
+    toggleCartView(event) {
+      store.commit("TOGGLE_VISIBILITY")
+    },
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+/*
+.circle-icon {
```
eliotjordan, line 196: question: This selector is commented out. Does we need to keep it?

app/javascript/components/RequestCart.vue
```diff
@@ -0,0 +1,522 @@
...
+  display: inline-flex;
+  align-self: center;
+  margin: 0 4px;
+}
+*/
+.lux-data-table {
+  /*display: block;*/
```
eliotjordan, line 203: issue: remove if not using

app/javascript/components/RequestCart.vue
```diff
@@ -0,0 +1,522 @@
...
+    font-family: franklin-gothic-urw, Helvetica, Arial, sans-serif;
+    line-height: 1;
+    &:last-child {
+      margin-bottom: 0;
+    }
+  }
+  tbody {
+    background-color: #fff;
+    width: 100%;
+    /* display: table; */
```
eliotjordan, line 227: issue: remove if not using

app/javascript/components/RequestCart.vue
```diff
@@ -0,0 +1,522 @@
...
+    background-color: #fff;
+    tr {
+      display: table-row;
+      /* display: block; */
```
eliotjordan, line 231: issue: remove if not using

app/javascript/components/RequestCart.vue
```diff
@@ -0,0 +1,522 @@
...
+.slide-leave-to {
+  transform: translateX(100%);
+  transition: 0.3s ease-out;
+}
+/* Component Styling */
+.request-cart {
+  padding-left: 0.8rem;
+  padding-right: 0.8rem;
+  /* Custom */
+  /* transform: translateX(0%); */
```
eliotjordan, line 407: issue: remove if not using

app/javascript/components/RequestCart.vue
```diff
@@ -0,0 +1,522 @@
...
+  top: 0px;
+}
+table {
+  margin-top: 2em;
+}
+.slide-enter-active,
+.slide-leave-active {
+  /*
+  transform: translateX(0%);
+  transition: 0.3s ease-out;
```
eliotjordan, line 499: issue: remove if not using

app/javascript/components/RequestCart.vue
```diff
@@ -0,0 +1,522 @@
...
+  */
+}
+.slide-enter,
+.slide-leave-to {
+  /*
+  transform: translateX(100%);
+  transition: 0.3s ease-out;
```
eliotjordan, line 506: issue: remove if not using

eliotjordan (changes requested):
I found a handful of minor issues, but otherwise this look great! Good work @tpendragon @jrgriffiniii!

tpendragon:
@sdellis @eliotjordan I think these are both taken care of.

eliotjordan (approved):

PR #448, opened 2020-11-02
Hide non-leaf node containers from requesting.

### LINKED ISSUE #441: "Add to Cart" should only appear on leaf node show pages.
We may be able to do this by either checking to see if a solr record has components or by checking to see if it has containers. The former is more structurally correct, the second relies on data entry best practices.

tpendragon:
Closes #441


jrgriffiniii (approved):
Testing against `aspace_WC064_c1` worked for me within the local development environment.

Issue #323, opened 2020-07-23
Integrate accessibility testing into the continuous integration workflow


jrgriffiniii:
Using axe-core-gems for rspec:

- https://github.com/dequelabs/axe-core-gems
- https://github.com/dequelabs/axe-core-gems/blob/develop/packages/axe-core-rspec/README.md

Pages to cover:
- [x] home
- [x] search results (I'll start on this - Anna)
- [x] collection summary / component show page (I'll start on this - Eliot)
- [x] request cart (#319)
- [x] expanded view (blocked until it exists)
- [ ] print view (blocked until it exists)

There are some forms that are ticketed separately.

Anything else from the wireframe? https://xd.adobe.com/view/a54bb217-10ff-43b2-72f6-6c3cadf376a4-d37a/

hackmastera:
This is built on axe-core and looks pretty easy to set up: https://thoughtbot.com/blog/introducing-accesslint-web-accessibility-testing-in-ci

hackmastera:
As I said on the PR, that accesslint tool doesn't render templates; it really only reviews html-formatted code.

This could be helpful if we want to try to run lighthouse: https://josephkhan.me/run-lighthouse-performance-audits/

hackmastera:
The axe-core tool integrates into javascript tests. We currently don't have any javascript tests, but even if we did they don't render our rails pages. The assumption of that tool is that you're running it against a javascript front-end.

I'm not sure we have a great option here.

hackmastera:
Okay we found the right axe gem that we can use within our feature tests (updated links in description). This seems like the easiest way forward.

The consequence of going this direction is that any failure will block merge because it will be an actual test failure.

The lighthouse integration would be harder to implement, but if we did move that direction at some point it would give us comments on PRs rather than test failures. I.e. it wouldn't block merge but it would give us useful feedback.

hackmastera:
with axe-core-rspec you have to tell it to run against wcag2a, wcag2aa, wcag21a, and wcag21aa to get all the rules. Here's the documentation: https://github.com/dequelabs/axe-core/blob/5e951536a6b4f21ef4d4a849b575b5a3c1ec6b85/doc/rule-descriptions.md

PR #450, opened 2020-11-03
Testing accesslint integration

hackmastera:
https://accesslint.com/pulibrary

refs #323


hackmastera:
So idk about this tool. here's a blog post about what it does / doesn't do: https://accesslint.com/blog/accesslint-github-marketplace-launch/

It seems like if we really want to do an accessibility CI check we'll need to actually run the site (i guess using docker somehow) as part of ci and run lighthouse on it.

PR #454, opened 2020-11-04
Component with diacritics

### LINKED ISSUE #282: Ensure diactrics render on site
This is an issue on the current site, particularly in the html view, that we'd like to make sure isn't replicated in pulfalight

christinach:
closes #282


christinach:
The fixture I use is from the finding aids with identifier C1408. One of the three that @faithc suggested in https://github.com/pulibrary/pulfalight/issues/282#issuecomment-721417785. 

During the checkin today we discussed that I should try to get a fixture from aspace because of the work that @tpendragon  is doing with the fixtures. I tried to find any of these 3 identifiers in https://asqa.lyrtech.org/staff/search?utf8=%E2%9C%93&q= or https://aspace.princeton.edu/staff/search?utf8=%E2%9C%93&q= but I was not able to.

christinach:
@tpendragon  I updated with the one from https://asqa.lyrtech.org/

lib/tasks/pulfalight.rake
```diff
@@ -168,6 +168,8 @@ namespace :pulfalight do
     # Delete previous fixtures. Needed for lando-based test solr.
     delete_by_query("<delete><query>*:*</query></delete>")
     index_directory(name: "spec/fixtures/ead/", root_path: Rails.root, enqueue: false)
+    puts "Seeding index with data from spec/fixtures/aspace..."
+    index_directory(name: "spec/fixtures/aspace/generated", root_path: Rails.root, enqueue: false)
```
christinach, line 172: I updated the existing rake task to index the generated eads from aspace instead of creating a new task.

tpendragon (approved):

PR #459, opened 2020-11-06
Use ReCAP as location when location code is rcp*

### LINKED ISSUE #447: Requests: Location should be ReCAP if site code is rcppa, rcpph, rcpxc, rcpxg, rcpxm, or rcpxr
See https://github.com/PrincetonUniversityLibrary/pulfa/blob/e8f383c4135948412fd0f0f88bfd73201f05e6a1/webapp/xsl/ead-classic-view.xsl#L241

Implement here: https://github.com/pulibrary/pulfalight/blob/0d3b5ac5287a465fdbf0569c637c603023b249cc/app/values/aeon_request.rb#L21

tpendragon:
Closes #447


hackmastera (commented):
So, not related to https://github.com/pulibrary/pulfalight/blob/5b2a9053b050e146c25c3a8f26b68004f8c3a85a/config/location_codes.yml? Why are we translating these values in 2 places?

tpendragon:
@hackartisan Ahh, yeah, that's a good point out. I think it's a little different here because we do want the "mss" code unless it's ReCAP. I dunno how best to resolve those two things, any suggestions?

eliotjordan (approved):

hackartisan:
@tpendragon sorry yeah I don't have a good sense of exactly what's going on in either place. maybe something worth writing an issue for? Should we try to put both in yaml files and call them 2 different things so at least the configuration is symmetrical?

PR #468, opened 2020-11-09
Don't index component physloc codes.

### LINKED ISSUE #451: Do not include physloc notes in "located in" field
From @faithc: The physloc text notes associated with container records, which note where in the stacks an item is located, should not display on the site. This is for staff use only. For example, see: https://pulfalight-staging.princeton.edu/catalog/aspace_C1491_c5621. "D-zone" should not display on the site.
Screen Shot 2020-11-04 at 8 36 42 AM

tpendragon:
Closes #451


christinach:
@tpendragon you need to rebase

tpendragon:
@christinach Thanks for the ping. Done.

christinach (approved):

PR #474, opened 2020-11-10
Add Collection-Level Storage Note

### LINKED ISSUE #225: Implement collection-level "Storage Note"
There are several notes whose labels and text strings are currently auto-generated via the production stylesheet for the current site (which @sdellis and @regineheberlein know about), including physloc type="code," which displays as the Storage note on the current site (https://findingaids.princeton.edu/collections/C0187/#summary), and prefercite, which displays as Credit this material on pulfa 2.0: https://findingaids.princeton.edu/collections/C0187/#accessuse. 

The Physical Location note on pulfalight is currently displaying the attribute value: https://pulfalight-staging.princeton.edu/catalog/C0187, which in this case is hsvm.

tpendragon:
Closes #225

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -316,6 +316,7 @@ class CatalogController < ApplicationController
     config.add_summary_field "title_ssm", label: "Title"
     config.add_summary_field "normalized_date_ssm", label: "Dates"
     config.add_summary_field "extent_ssm", label: "Size"
+    config.add_summary_field "storage_note_ssm", label: "Storage Note", accessor: :fetch_html_safe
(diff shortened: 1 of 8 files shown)


hackartisan (changes requested):
It seems good, but I don't really understand why we want to run the field through a presenter as part of indexing. Wouldn't it make more sense to compute the value at display time, so that when these notes change we don't have to reindex everything?

tpendragon:
Yup, good call, not sure what I was thinking. Thanks!

tpendragon:
@hackartisan Should be good for re-review.

hackartisan (approved):

hackartisan:
Thanks!

PR #476, opened 2020-11-10
Add "Suggest a Correction"

### LINKED ISSUE #105: Users should be able to submit corrections for content in a collection or component
Wireframe: https://xd.adobe.com/view/a54bb217-10ff-43b2-72f6-6c3cadf376a4-d37a/screen/8511ff6c-e5b8-4a41-9926-9c90325d315a

This is similar to #104, and is also currently supported with PULFA 2.0 with a "Suggest a correction" link:

image

image

tpendragon:
Closes #105

Screen Shot 2020-11-10 at 1 06 19 PM

Screen Shot 2020-11-10 at 1 06 26 PM

Screen Shot 2020-11-10 at 1 06 34 PM


.rubocop.yml
```diff
@@ -54,24 +56,14 @@ Metrics/PerceivedComplexity:
...
-    - 'spec/models/solr_document_spec.rb'
-    - 'spec/requests/catalog_request_spec.rb'
-    - 'spec/services/robots_generator_service_spec.rb'
-    - 'spec/services/table_of_contents_builder_spec.rb'
-    - 'spec/jobs/aspace_index_job_spec.rb'
-    - 'spec/services/aspace/client_spec.rb'
-    - 'spec/services/aspace/indexer_spec.rb'
-    - 'spec/services/aspace_fixture_generator_spec.rb'
-    - 'spec/features/accessibility_spec.rb'
+  Enabled: false
```
hackartisan, line 59: 🎉

hackartisan (changes requested):
this is a little bit more complicated than I had time to fully review so if no one else looks at it I can look more in the morning. But initially I was wondering about the flash message in the modal. I think it might be better to close the form on submission and show the flash on the record page, did/would you consider doing it that way?

tpendragon:
@hackartisan I did try it with the flash in the main page, the problem is to get to suggest a correction you're probably scrolled down, so you click the thing, it closes out, and you can't see the flash message which has appeared but above the cut. I could scroll them up but it seems unnecessarily disruptive.

hackartisan:
Hm, that's interesting. And the button is all the way at the bottom of the page?

tpendragon:
@hackartisan It's below the descriptive metadata.

Screen Shot 2020-11-10 at 1 42 56 PM

hackartisan:
What about clearing the modal and just having the flash message then? I think it is sort of confusing to present the form again.

tpendragon:
@hackartisan I left the form so they wouldn't have to refresh the page again to submit another request, and I don't have to dynamically load the form on clicking the button.

hackartisan:
Do people want to submit more than one correction by filling out the form multiple times? That would be surprising to me.

tpendragon:
@hackartisan I'm not sure, but they probably don't want to close out of the modal and get thanked again until they refresh. I can make it load the form dynamically on clicking the button if this is a blocker, I personally think the way it works now is pretty nice.

hackartisan:
Okay. I don't really like it but it seems like it's up to the POs and I won't block on it.

Unfortunately I wasn't able to give this pr more than a very quick look. It seems good but I wouldn't want to approve without a closer review. if someone else reviews it feel free to override my block to merge.

btw, thank you for all the screenshots!

tpendragon:
@hackartisan Went ahead with your suggestion:

Screen Shot 2020-11-10 at 3 20 20 PM

It just hides the form with JS, and displays it again when you hit the button.

tpendragon:
Don't merge this please, I think the location codes being used are wrong but I'm done for the night.

app/views/layouts/blacklight/base.html.erb
```diff
@@ -41,5 +41,6 @@
     <noscript><%= render partial: 'shared/footer_noscript' %></noscript>
     <%= render partial: 'shared/footer' %>
     <%= render partial: 'shared/modal' %>
+    <%= content_for(:correction_modal) %>
```
hackartisan, line 44: How does this work? Does it not have to be `yield`ed?

Interesting to place it with the layout but render it from the show pages as a `content_for` block. I'm just not sure how this bit works.

lib/templates/erb/scaffold/_form.html.erb
```diff
@@ -0,0 +1,15 @@
+<%# frozen_string_literal: true %>
```
hackartisan, line 1: was this added by simple form? Or does this overwrite some simpleform functionality? Maybe add a comment to this file describing what it's all about.

spec/mailers/contact_mailer_spec.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe ContactMailer, type: :mailer do
+  pending "add some examples to (or delete) #{__FILE__}"
```
hackartisan, line 5: Do you want to add a test or delete this?

spec/forms/suggest_a_correction_form_spec.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.describe SuggestACorrectionForm do
+  let(:valid_attributes) do
+    {
+      "name" => "Test",
+      "email" => "test@test.org",
+      "box_number" => "1",
+      "message" => "This is so broken.",
```
hackartisan, line 10: 😆 I got confused for a minute because the valid attribute says it's broken.

hackartisan (changes requested):
I think it looks nice this way!

Wow, really nice job on this PR. I added a couple of questions and I'm also blocking based on your comment about the locations.

app/views/layouts/blacklight/base.html.erb
tpendragon: It's in the layout because it needs to be in a specific spot in the DOM or else the modal doesn't work. I use `content_for` because I don't want the form to always be rendered.

Documentation for `content_for` is here: https://guides.rubyonrails.org/layouts_and_rendering.html#using-the-content-for-method

hackartisan: Yes I'm with you and in general the approach makes sense, but the `content_for` is what you use to wrap the content. Doesn't it need to be yielded? like see above in line 36. I assume it works, though. maybe we could look at it together.

tpendragon: Oh, yeah, `content_for` with a name and a block is equivalent to yielding to a symbol in the view. See https://github.com/pulibrary/pulfalight/pull/476#discussion_r521556966.

hackartisan: > Oh, yeah, content_for with a name and a block is equivalent to yielding to a symbol in the view. See https://api.rubyonrails.org/classes/ActionView/Helpers/CaptureHelper.html#method-i-content_for.

Aha! you mean "with a name and without a block" I think.

So we could rewrite line 36 this way? (Not proposing to do that on this pr just hypothetically.)

tpendragon: Yes

hackartisan (approved):

PR #479, opened 2020-11-11
389 query figgy viewer

### LINKED ISSUE #389: Query Figgy for viewer if there's no DAO
Split from https://github.com/pulibrary/pulfalight/issues/364

Similar to https://github.com/PrincetonUniversityLibrary/pulfa/pull/1225

I split this issue out because it was hard to find an example resource to test this with right now. When we have more data to index it'll be easier.

### LINKED ISSUE #511: Append style in the readingroom div when there is a successful response from checkFiggy.
Currently we add the div in the DOM when the page loads. We should first check that the response is successful and then append the div.
blocked by #389

christinach:
closes #511
closes #389


christinach:
Screen Shot 2020-11-11 at 6 06 43 PM

christinach:
After login in

snapshot_after_login

app/javascript/pulfalight/pulfalight_loader.es6
```diff
@@ -9,13 +9,23 @@ import AddToCartButton from '../components/AddToCartButton'
...
     this.setup_lib_cal_hours()
     this.setup_range_limit()
+    this.query_figgy()
+  }
+
+  query_figgy(){
+    const query = new QueryFiggy()
+    let doc_aspace_component_id = document.getElementById('document').getElementsByTagName('div')[0]
+    let component_id = doc_aspace_component_id.getAttribute('id').replace('doc_aspace_', '')
+    component_id = component_id.charAt(0).toUpperCase() + component_id.slice(1)
```
hackartisan, line 27: Only the first char should be uppercase? Shouldn't any leading alphabetic characters be uppercased? I might move this line into `QueryFiggy` so it can be tested.

hackartisan (commented):
Are we planning to have tests here? Not sure if this is ready for review.

I have a question regardless about the "view content" button. Not sure what it links to and whether it should be visible in these screen shots. (thank you for attaching screen shots!)

app/javascript/pulfalight/pulfalight_loader.es6
christinach: Do the component_ids in figgy have more than one leading alphabetic character? If yes and if they are upper-cased then `any leading alphabetic character` should be uppercased. I can check on that as well.

christinach:
> Are we planning to have tests here? Not sure if this is ready for review.
> 
> I have a question regardless about the "view content" button. Not sure what it links to and whether it should be visible in these screen shots. (thank you for attaching screen shots!)

I'm not sure about the view content button. I think this is a different issue. There is one test to make sure that QueryFiggy is defined and another test to make sure the login button is there. I can add more jest tests but I need more time.

app/javascript/pulfalight/pulfalight_loader.es6
tpendragon: Yeah, for example `AC101`

christinach: ok, I'll update. thanks

hackartisan (changes requested):
We would like to add some space between the access message box and the button; this review is a blocker for that work!

tpendragon (approved):

tpendragon (approved):

app/javascript/pulfalight/query_figgy.es6
```diff
@@ -0,0 +1,118 @@
+export default class QueryFiggy {
```
hackartisan, line 1: I think this should this just be `export default class` without a class name

spec/javascript/pulfalight/query_figgy.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    const query = new QueryFiggy
+    await query.checkFiggy(component_id);
+    expect(mockSuccessXHR.send).toHaveBeenCalled();
+    expect(document.querySelector('#readingroom iframe').getAttribute('src')).toBe('https://figgy.princeton.edu/viewer#?manifest=https://figgy.princeton.edu/concern/scanned_resources/78e15d09-3a79-4057-b358-4fde3d884bbb/manifest')
+  })
+
+  test("component_id() Capitalizes alphabetic characters before the underscore", () => {
+    document.body.innerHTML = '<div id="document" class="document blacklight-file" itemscope="" itemtype="http://schema.org/Thing"><div id="doc_aspace_c1491_c4"></div></div>'
+    const query = new QueryFiggy
+    expect(query.component_id(component_id)).toEqual("C1491_c4")
```
hackartisan, line 66: 👍

hackartisan (changes requested):

hackartisan (approved):

PR #512, opened 2020-11-19
Reconfigure access facet, push to the top, and make always open

### LINKED ISSUE #232: Rename "Online Access" Facet to "Access", put it at the top, and make it always open.
Technically, there is only one other way to access the materials and that's in person, which there is no toggle option for. It seems unnecessary for researchers to have to click the menu to click the option to see the refreshed results. If the drop down menu needs to be there, then should the rest of the materials be labeled with "In the Library" access like in the Catalog? (see screenshots of staging, current, and catalog examples).
staging site: Screen Shot 2020-04-07 at 10 47 42 AM
current site: Screen Shot 2020-04-07 at 10 52 38 AM
catalog example: Screen Shot 2020-04-07 at 10 54 32 AM

eliotjordan:
Closes #232 

Screen Shot 2020-11-19 at 5 31 23 PM

### DIFF
--- a/spec/features/facets_spec.rb
+++ b/spec/features/facets_spec.rb
@@ -20,4 +20,17 @@ describe "faceted searches", type: :feature, js: true do
       expect(page).to have_css("#facet-date_range_sim .missing", text: /Unknown/, visible: false)
     end
   end
+
+  describe "access facet" do
+    before do
+      visit "/?search_field=all_fields&q="
+    end
+
+    it "displays access values and is always open" do
+      expect(page).to have_selector("h3.facet-field-heading:first-child button", text: "Access", visible: false)
+      expect(page).to have_selector("#facet-has_online_content_ssim.show", visible: false)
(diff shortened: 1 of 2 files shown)


spec/features/facets_spec.rb
```diff
@@ -20,4 +20,16 @@
       expect(page).to have_css("#facet-date_range_sim .missing", text: /Unknown/, visible: false)
     end
   end
+
```
sdellis, line 23: @eliotjordan does this test the new order? We probably want to check the text of `h3.facet-field-heading:first-child button` is "Access", right?

sdellis (changes requested):

spec/features/facets_spec.rb
eliotjordan: Yes. Good catch.

eliotjordan:
@sdellis Updated.

sdellis (approved):
🎉

PR #524, opened 2020-11-25
Improve empty request cart behavior

### LINKED ISSUE #504: Cart has no 'x' to close when empty
It has an 'x' to close when it has contents. It should be consistent.

The heading is also a different color (black instead of grey). Break this out into a separate issue if it doesn't make sense to do both together.

eliotjordan:
- Most importantly, ensures that there is a closed button when the cart is empty.
- Adds feature tests for request cart

Closes #504 

Screen Shot 2020-11-25 at 9 28 54 AM

### DIFF
--- a/app/javascript/components/RequestCart.vue
+++ b/app/javascript/components/RequestCart.vue
@@ -93,7 +93,7 @@
         </tbody>
       </table>
 
-    </div>
+    </div><!-- /.panel -->
 
     <div class="cart-actions">
       <div class="center">
--- a/app/javascript/components/RequestCart.vue
+++ b/app/javascript/components/RequestCart.vue
@@ -4,7 +4,7 @@
 
   <div v-if="isVisible" :class="['request-cart']">
 
-    <div v-if="requests.length" class="panel">
+    <div class="panel">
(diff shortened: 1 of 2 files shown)


app/javascript/components/RequestCart.vue
```diff
@@ -93,9 +93,6 @@
         </tbody>
       </table>
 
-    </div><!-- /.panel -->
```
hackartisan, line 96: maybe keep the div with the comment and delete the other one? It's sort of helpful since it's very far away from its opening tag.

hackartisan (changes requested):

app/javascript/components/RequestCart.vue
```diff
@@ -93,9 +93,6 @@
...
-    <div v-else class="panel">
-      <heading level="h3">Your cart is currently empty.</heading>
     </div>
```
tpendragon, line 96:
```suggestion
    </div><!-- /.panel -->
```

tpendragon (changes requested):

tpendragon (approved):

hackartisan (approved):

PR #528, opened 2020-11-25
Remove child documents from document query

eliotjordan:
Reduces the load time for large EAD collection pages by a huge amount. On staging, the Toni Morrison collection page went from a 20 second load time to 260ms. I don't think we need to return all of the nested component children for any show page functionality.


eliotjordan:
On second thought, I see there is functionality to render a minimal document for HTML for AJAX requests? Is that more important than intolerable load times? In fact, this seem obsolete as it was primarily intended for a table of contents or expanded collection view, neither of which use this functionality.

https://github.com/pulibrary/pulfalight/blob/main/spec/requests/catalog_request_spec.rb#L37
https://github.com/pulibrary/pulfalight/issues/239
https://github.com/pulibrary/pulfalight/pull/243

tpendragon:
@eliotjordan Yeah I think that was for the old "collection context", and we have the table of contents now.

eliotjordan:
I was able to reduce the expanded view render time from 20 seconds to 5 five seconds by limiting the child fields the solr document tree requests.

tpendragon (approved):

PR #602, opened 2021-01-22
Modifying the styling for the catalog show view in order to ensure that the request cart is responsive for mobile devices

### LINKED ISSUE #440: Request Cart should scale for Mobile
Right now it's too small to use, and the button doesn't appear.

jrgriffiniii:
Resolves #440 and providing a solution for a small bug in the Rake tasks


tpendragon:
@jrgriffiniii This is good progress. I'm going to merge this, but it looks like you still can't access the request cart button in mobile view in the header, right?

tpendragon (approved):

PR #607, opened 2021-01-25
Adds Ask A Question button to top level of collection

### LINKED ISSUE #583: Ask a Question button should display in Collection navigation
The ask a question button is currently not included in the bottom navigation bar on collection-level description pages (the pages/tabs that are part of the top navigation bar: Overview, Collection Description, Collection History, etc.), for example: https://findingaids-beta.princeton.edu/catalog/C1491. It should appear on these pages as well.

sdellis:
Fixes #583

### DIFF
--- a/app/views/catalog/_show_collection.html.erb
+++ b/app/views/catalog/_show_collection.html.erb
@@ -56,14 +56,14 @@
             <div class="row">
               <div class="col-md-4"></div>
               <div class="col-md-4">
-                <a href="#question" data-toggle="modal" data-target="#questionModal">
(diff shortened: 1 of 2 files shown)


tpendragon (changes requested):
@sdellis Looks great! Can you add a test? Probably something like https://github.com/pulibrary/pulfalight/blob/d9365acada495c64ff7518915489b721d3943375/spec/features/catalog_view_spec.rb#L22-L29 but with `MC221`

tpendragon (approved):

PR #621, opened 2021-01-27
Rename tab headers.

### LINKED ISSUE #585: Top navigation bar headings
The headings in the top navigation bar for finding aids should be as follows: 

Overview | Collection Description & Creator Information | Access & Use | Collection History | Find Related Materials

tpendragon:
Closes #585

### DIFF
--- a/app/views/catalog/_show_collection.html.erb
+++ b/app/views/catalog/_show_collection.html.erb
@@ -15,17 +15,17 @@
           </li>
           <li class='nav-item flex-fill' role='tab'>
             <a class='nav-link ' data-toggle='tab' href='#description' role='tab'>
-              Collection Description
-            </a>
-          </li>
-          <li class='nav-item flex-fill' role='tab'>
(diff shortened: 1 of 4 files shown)


app/views/catalog/_show_collection.html.erb
```diff
@@ -15,17 +15,17 @@
           </li>
           <li class='nav-item flex-fill' role='tab'>
             <a class='nav-link ' data-toggle='tab' href='#description' role='tab'>
-              Collection Description
+              Collection Description & Creator Information
```
jrgriffiniii, line 18: Should this instead use `&amp;` or `&#x00026;`?

jrgriffiniii (approved):

app/views/catalog/_show_collection.html.erb
tpendragon: I don't think you need to escape it just for HTML display do you?

PR #624, opened 2021-01-28
mostly fixes bottom action bar

sdellis:


tpendragon (changes requested):
It needs rebased, but also when there's long content the scroll float seems a little off now.

See https://pulfalight-staging.princeton.edu/catalog/aspace_MC001-01_c002356. When you scroll half the actions pane is cut off.

tpendragon (approved):

PR #640, opened 2021-02-02
Table of contents opens direct children when clicked on.

### LINKED ISSUE #576: Arrows in ToC are not intuitive
We have had many comments that it is not intuitive that you have to click on the arrow itself in the ToC to expand. Many users just want to click on the title next to the arrow to expand. 

# Implementation Note

When navigating to a record the ToC should expand its children.

tpendragon:
Closes #576

ToC looks like this when navigating directly to a component now, expanding its direct children:

Screen Shot 2021-02-02 at 1 57 26 PM

### DIFF
--- a/app/services/table_of_contents_builder.rb
+++ b/app/services/table_of_contents_builder.rb
@@ -19,14 +19,14 @@ class TableOfContentsBuilder
 
   private
 
-  def build_tree(solr_docs)
+  def build_tree(solr_docs, level: 0, full: false)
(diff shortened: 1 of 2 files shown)


jrgriffiniii (approved):

ccleeton:
So the items open when clicked on but they do not close when clicked on. So if I am in a subseries and see that its not the one I need I can't click the TOC to close it again.

PR #643, opened 2021-02-03
Don't always display digital content tag.

### LINKED ISSUE #642: Has Online Content tag displaying in error
I noticed that the tag "Has Online Content" is displaying for this collection even though there's no dao: https://findingaids-beta.princeton.edu/catalog/C0879. Not sure how widespread this issue is?

tpendragon:
Closes #642

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -221,7 +221,7 @@ class SolrDocument
   end
 
   def has_digital_content?
-    has_online_content.present?
+    has_online_content.present? && has_online_content.first != "false"
   end
 
   def component_documents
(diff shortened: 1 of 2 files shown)


jrgriffiniii (approved):

app/models/solr_document.rb
```diff
@@ -221,7 +221,7 @@ def component?
   end
 
   def has_digital_content?
-    has_online_content.present?
+    has_online_content.present? && has_online_content.first != "false"
```
jrgriffiniii, line 224: I am sorry, I was not aware that `false` this was a possible value for the first element in this array.

PR #677, opened 2021-02-22
Makes the Collection tabs into links so we can show Collection Summary info

sdellis:


app/assets/javascripts/tabs.js
```diff
@@ -24,36 +24,45 @@ $( document ).ready(function() {
...
-    }
+  $(".lux-tabs-container li").click(function() {
+    let temp_id = $(this).attr('id').split("-");
+    temp_id.pop()
+    window.location.hash = temp_id.join("-");
+    console.log($(this).attr('id'))
+    window.location.reload()
   });
+
+  // $(".lux-tabs-container li").click(function( event ) {
```
tpendragon, line 38: Can you delete this instead of commenting?

app/views/catalog/_show_tabs.html.erb
```diff
@@ -1,26 +1,26 @@
 <ul class="lux-tabs-container" role="tablist" aria-label="Collection Notes">
   <li id="summary-tab" class="lux-tabs-item" role='tab'>
-    <a class='nav-link' href='#summary'>
-      Collection Overview
+    <a class='nav-link' href='<%= solr_document_path(document.http_safe_ead) %>'>
```
tpendragon, line 3: What's `http_safe_ead`? Did `document.id` not work?

app/assets/javascripts/tabs.js
```diff
@@ -1,6 +1,6 @@
 $( document ).ready(function() {
   const url = new URL(window.location);
-
+  $(".lux-tabs-container li").removeClass( "active" );
```
tpendragon, line 3: A weird thing happens when I press "back" in the browser - the anchor sticks around as #access but on the component. Maybe a fluke?

tpendragon (changes requested):

app/views/catalog/_show_tabs.html.erb
sdellis: I needed the Collection URL, and that was the way I found I could get it. The Document.id is the component URL if you're on a component.

app/assets/javascripts/tabs.js
sdellis: @tpendragon I think I fixed this in my latest push. LMK if that works for you.

app/assets/javascripts/tabs.js
sdellis: Yes. I added a permalink to the old code for future reference.

app/assets/javascripts/tabs.js
tpendragon: @sdellis The back button thing still happens for me. https://pulfalight-staging.princeton.edu/catalog/C0159_c000001 Click "access and use" and then hit the back button. Maybe we'll have to just not render those tabs at all unless you're on a collection component?

sdellis: @tpendragon I think I finally got it with this latest commit.🤞  Thanks for catching that.

tpendragon: Yup, that seems to work. PR just needs rebased tomorrow and we're good.

tpendragon (approved):

Issue #574, opened 2020-12-14
Series/component-level notes are not displaying


faithc:
This is likely being addressed as part of the indexing work, but just wanted to note that component-level notes aren't displaying, only collection-level notes. Ex: https://findingaids-beta.princeton.edu/catalog/aspace_C1619_c24. 

This also includes missing origination and bioghists at series/component levels, ex. https://findingaids-beta.princeton.edu/catalog/aspace_C0033_c001

Relates to #279

tpendragon:
I have origination and scopecontent working in #616. I can't do bioghist yet because the EADs that get exported from ASpace don't include the component's bioghist, just the collection's.

tpendragon:
@faithc It might be a lot to ask, but if there are other series notes could you find examples like you did for those three? It helps a lot.

faithc:
Another example: https://findingaids-beta.princeton.edu/catalog/C1596; none of the series/component-level notes are displaying, only some of the Collection overview notes. 
https://findingaids-beta.princeton.edu/catalog/C1596_c1 vs. https://findingaids.princeton.edu/collections/C1596/c1
https://findingaids-beta.princeton.edu/catalog/C1596_c78 vs. https://findingaids.princeton.edu/collections/C1596/c78

tpendragon:
@faithc I think I've got this locked up. Can you check again?

faithc:
Hmm, there are no paragraph breaks here either and some wonky formatting.

tpendragon:
In which fields? It might be the source data no longer has the breaks

faithc:
All of the scopecontent notes.

faithc:
But yes, the data is now displaying. :)

faithc:
Looks like ASpace does have the paragraph breaks. I just checked.

tpendragon:
That's an easy fix - https://findingaids-beta.princeton.edu/catalog/C1596_c1 should be better now.

faithc:
http://gph.is/1OXRUog

tpendragon:
@faithc Safe to close this?

faithc:
Yes, I think so. Thanks.

faithc:
[...]

PR #682, opened 2021-02-23
Add more metadata to component display.

tpendragon:
Work towards #574

This removes some of the data supporting the tabs. This is because those tabs are now going to become links (see #677). In the future if we want them be tabs again we should populate the content from the collection document rather than reindexing all the parent metadata in the children.

Screen Shot 2021-02-23 at 11 35 00 AM


app/controllers/catalog_controller.rb
```diff
@@ -337,15 +337,46 @@ def item_requestable?(_ = nil, _options = {})
...
+    config.add_component_field "names_coll_ssim", label: "Names", separator_options: {
+      words_connector: "<br/>",
+      two_words_connector: "<br/>",
+      last_word_connector: "<br/>"
+    }, helper_method: :link_to_name_facet
+
+    config.add_component_field "places_ssim", label: "Places", link_to_facet: true, separator_options: {
+      words_connector: "<br/>",
+      two_words_connector: "<br/>",
+      last_word_connector: "<br/>"
```
jrgriffiniii, line 377: I was not aware of this pattern, 👍

jrgriffiniii (approved):

PR #687, opened 2021-02-24
Namespace setup_keys task under pulfalight to match other tasks

bess:
This will also make it match the instructions in the README


tpendragon (changes requested):
Can we change the README instead? We have other projects with a `setup_keys` and we should probably standardize what the task is called.

PR #688, opened 2021-02-24
Fix README to match expected syntax for setup_keys task

bess:


tpendragon (approved):
Thanks! Sorry it was wrong. 😅

PR #695, opened 2021-02-24
Components return on match of collection title.

### LINKED ISSUE #693: Components should return if their collection title matches.
https://findingaids-beta.princeton.edu/catalog?q=World+press+freedom+committee+records&search_field=all_fields should return all the results of World Press Freedom Committee Records so that faceting by online content after a keyword search works.

tpendragon:
Closes #693

### DIFF
--- a/lib/pulfalight/traject/ead2_component_config.rb
+++ b/lib/pulfalight/traject/ead2_component_config.rb
@@ -95,6 +95,13 @@ to_field "parent_ssm" do |record, accumulator, context|
   end
 end
 
+to_field "collection_title_tesim" do |_record, accumulator, context|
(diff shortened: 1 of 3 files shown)


lib/pulfalight/traject/ead2_component_config.rb
```diff
@@ -95,6 +95,13 @@
   end
 end
 
+to_field "collection_title_tesim" do |_record, accumulator, context|
+  parent = context.clipboard[:parent] || settings[:root]
```
jrgriffiniii, line 99: I understand `settings[:root]` here to reference the top of the component tree within the description 👍

jrgriffiniii (approved):

Issue #661, opened 2021-02-10
Landing page image


ccleeton:
Could the landing page image be changed to this photo - https://drive.google.com/file/d/1XbiRLXORUHx2dpCOBVC89ej_iqZPCGJn/view
If possible we would prefer to not have the students covered by the search bar.

Would it be possible to do this on the staging site first so we can get leadership approval?

tpendragon:
@ccleeton Blocking release?

ccleeton:
I would say yes since the image has been approved by Leadership and Communications. So as long as its done by March??

bess:
@ccleeton Please take a look at https://pulfalight-staging.princeton.edu/ when you get a chance. I'm also pasting a screenshot. What is the process for getting sign off on this?  Thanks!

![Screen Shot 2021-03-01 at 9.59.54 AM.png](https://images.zenhubusercontent.com/5ca3a84155cfe32f2f66b74c/9d743319-b11c-42ea-b2fe-e9990e98001a)

ccleeton:
Looks good to me. Let me see if I can get approval on this quickly.

ccleeton:
@bess Approved from Sara and Alexis. This can be pushed to the beta site whenever you can.

bess:
Great, thank you, @ccleeton !

bess:
Merged and deployed to beta site!

PR #702, opened 2021-02-25
Change image on front page

bess:
Use the image of Firestone special collections from ticket #661
Position so as much as possible the students are not covered by
the search box.

Here are some screen shots at different widths:

Screen Shot 2021-02-25 at 2 17 14 PM

Screen Shot 2021-02-25 at 2 17 51 PM


tpendragon:
I'm pretty surprised this passes the accessibility tests. Seems like we need some way to better see the title text. Maybe a darkened opacity on the image? Any suggestions @sdellis?

bess:
Good point. I'll try adding some opacity to the image.

bess:
Here it is with an opacity of .3 on the image, and I swapped the h1 color to be $black-pul. 

Screen Shot 2021-02-25 at 3 32 02 PM

ccleeton:
I like this with the opacity!

tpendragon:
I find it pretty hard to read the text with so much going on behind it, but I have no good suggestions for how to improve that. It might just be a personal problem too.

sdellis:
@tpendragon I think that darkening the image, keeping the white lettering, and adding a drop shadow should do the trick. I'm happy to give it a try if it helps.

bess:
@sdellis Would you please give that a shot? It sounds like your UI skills vastly exceed mine. Thank you!

sdellis:
HOMEPAGE WITH DROPSHADOW

Here's a version with a darkened image, dropshadow, and centering of text.

sdellis:
FYI - axe is not taking the shadow into consideration in its test. This is perfectly acceptable with regard to accessibility though. See: https://www.w3.org/TR/WCAG20-TECHS/G18.html

> If the background or the letters vary in relative luminance (or are patterned) then the background around the letters can be chosen or shaded so that the letters maintain a 4.5:1 contrast ratio with the background behind them even if they do not have that contrast ratio with the entire background.

tpendragon:
@sdellis @bess Just need rubocop run, I think.

tpendragon (approved):

PR #703, opened 2021-02-25
Group by a unique identifier for requests.

### LINKED ISSUE #696: Collections merging in call slip when the box number and component is same
We have had a few instances of the merging of requests into one transaction when the collection information is different. This is seeming to happen when the box is the same and the component is the same but from different collection.
Example: Requesting AC109_c00003; AC168_c00003
This is an issue because we need separate call slips for each collection. 
merging collections

tpendragon:
Adding a grouping identifier for Aeon which includes the EADID ensures
that only the same box merges into the same call slip.

Closes #696

Special thanks to https://blog.rockarch.org/technical-overview-of-ead-requesting-with-aeon and @helrond (which is funny, because PULFA2 is referenced there. :joy:)

### DIFF
--- a/spec/models/solr_document_spec.rb
+++ b/spec/models/solr_document_spec.rb
@@ -255,7 +255,7 @@ RSpec.describe Arclight::SolrDocument do
       expect(request.form_attributes[:Site]).to eq "RBSC"
       expect(request.form_attributes[:Location]).to eq "mss"
       expect(request.form_attributes[:ItemTitle]).to eq "Diary"
-      expect(request.form_attributes[:GroupingIdentifier]).to eq "ItemVolume"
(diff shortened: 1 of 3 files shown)


jrgriffiniii (approved):

spec/models/solr_document_spec.rb
```diff
@@ -268,6 +268,8 @@
       # https://findingaids.princeton.edu/collections/C1588/c2
       expect(request.form_attributes[:Request]).not_to be_blank
       request_id = request.form_attributes[:Request]
+      # Ensure the box/collection unitid are in as a grouping identifier.
+      expect(request.form_attributes[:"GroupingField_#{request_id}"]).to eq "C1588test-Box B-001180"
```
jrgriffiniii, line 272: 👍

PR #715, opened 2021-03-02
Display component list in show page.

### LINKED ISSUE #573: Child Components should be requestable and sortable from their parent's show screen
When viewing components with children, users need to be able to sort child components by title, date, and container. They should also be able to request child components (end-node, without children) from this screen simply by adding them to their cart. See examples below:

Wireframes ([URL](https://xd.adobe.com/view/a54bb217-10ff-43b2-72f6-6c3cadf376a4-d37a/screen/2aeefae6-aee7-4ec5-929a-64f600503129)):
Screen Shot 2020-12-14 at 4 47 06 PM

Current PULFA site ([URL](https://findingaids.princeton.edu/collections/C0108/c00584)) :
Screen Shot 2020-12-14 at 4 48 02 PM

tpendragon:
Closes #573

Screen Shot 2021-03-04 at 3 36 46 PM

Screen Shot 2021-03-04 at 3 36 58 PM


sdellis:
@tpendragon I think this is ready for review, tests, and/or further discussion. I had to remove the ChildTable from the Collection Overview tab pane as it was rendering the table twice in the DOM and creating id collisions. (Pulfa 2.0 does not have this on the Collection Overview screen.)

tpendragon:
This is great work! A small bug I noticed is container info doesn't seem to be propagating via the button. I need to add some tests to the builder, do you want me to pick up from here?

tpendragon:
Converted back to draft - we're going to have to do this via some sort of ajax for cases like https://pulfalight-staging.princeton.edu/catalog/RBD1-1_c18905, where there's hundreds of or thousands of child components.

tpendragon:
This ajax method works, but loading the one above takes -several- minutes. I'm gonna call it a night, we should talk about this.

tpendragon:
We're gonna leave this the way it is and improve performance later.

PR #738, opened 2021-03-10
Add a patch for blank boxes before reindexing is finished.

tpendragon:


app/values/aeon_request.rb
```diff
@@ -33,6 +33,7 @@ def all_request_attributes
     all_attributes = boxes.map do |local_box|
       request_attributes(local_box)
     end
+    return {} if all_attributes.blank?
```
jrgriffiniii, line 36: I am sorry for missing this, I would not have assumed that an empty `[]` would need to be `{}` for this case.

jrgriffiniii (approved):

PR #742, opened 2021-03-10
Revert "Group search results by collection by default"

### LINKED ISSUE #113: Revert "Group search results by collection as a default (rather than the "All results" view)"

tpendragon:
This reverts commit c6bb6ef696962c113d416cfe2c8250a07b0651ef.

Closes #113


tpendragon:
Marking this as Draft because we may be keeping the default as grouped by collection.

PR #752, opened 2021-03-12
Removing coveralls badge, adding license and related badge

escowles:


tpendragon (approved):

LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
tpendragon, line 190: I think we need to replace this.

tpendragon (changes requested):

LICENSE
escowles: We have two different styles: in some places we are using the whole license file verbatim like this, in other places, we are just putting the copyright notice and link to the license (e.g., in Figgy: https://github.com/pulibrary/figgy/blob/main/LICENSE).

I think the Figgy is probably the better way, so I can change this to match that

tpendragon (approved):

PR #765, opened 2021-03-16
Add Note to request cart.

### LINKED ISSUE #671: Notes field when requesting
There is not a notes field that is available to researchers when placing a request. Researchers and staff use this often. Could a sub screen appear after clicking request from the cart and before logging into aeon. Similar to the way 2.0 works now.

tpendragon:
* Adds a note field to the request cart.
* Adjusts CSS for request cart to allow for scroll no matter how many
  components there are.
  - This uses CSS3, but works in all major browsers.
  
  Closes #671

### DIFF
--- a/app/javascript/components/RequestCart.vue
+++ b/app/javascript/components/RequestCart.vue
@@ -123,7 +123,7 @@
               label="Notes"
               name="Notes"
               :hide-label="true"
-              placeholder="Notes"
+              placeholder="Notes to Special Collections Staff"
               width="expand"
               type="textarea"
               rows=2
--- a/app/javascript/components/RequestCart.vue
+++ b/app/javascript/components/RequestCart.vue
@@ -16,7 +16,6 @@
             variation="text"
             class="denied-button"
             aria-labelledby="closeCart"
-            focused
             >
 
             <div class="lux-icon">
@@ -117,6 +116,19 @@
             </template>
           </div>
 
+          <div id="request-notes-wrapper">
+            <input-text
+              id="request-notes"
+              label="Notes"
+              name="Notes"
+              :hide-label="true"
+              placeholder="Notes"
+              width="expand"
+              type="textarea"
+              rows=2
+              v-model="note"
+            />
+          </div>
           <input-button type="submit" variation="solid" :disabled="requests.length == 0" block>
             {{ requestButtonText() }}
           </input-button>
@@ -139,6 +151,7 @@
                                                               :value="form_values"></input>
           </template>
         </template>
+        <input name="Notes" type="hidden" :value="note" />
       </div>
     </form>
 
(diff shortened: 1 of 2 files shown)


tpendragon:
Screen Shot 2021-03-16 at 11 47 10 AM

ccleeton:
I like this, looks like the note will add to each item in the cart when submitted. Not sure if that is indeed what we need/want. Let me run this by PS during our wednesday meeting.

app/javascript/components/RequestCart.vue
```diff
@@ -16,7 +16,6 @@
             variation="text"
             class="denied-button"
             aria-labelledby="closeCart"
-            focused
```
sdellis, line 19: Unless there's a good reason to focus this field, I think it's best to apply the focus as it was, at the top of the panel. This is because it may not be obvious to a user with a screen reader that the focus is in the middle of the panel.

sdellis (changes requested):
Great work, @tpendragon! I only have one reservation about the auto-focusing of the notes field. Other than that this is 👍.

app/javascript/components/RequestCart.vue
tpendragon: I had to get rid of this because with it around it kept re-rendering the template on unfocusing the note field, causing it to wipe out whatever was typed.

sdellis: @tpendragon, I see. One solution might be to store the note in the vuex store on keydown to prevent that. I will approve this as is and attempt to solve that focus issue in another ticket.

sdellis (approved):

ccleeton:
I will say if you want to deploy it would seem to be fine and then we can refine if needed.

tpendragon:
There's no rush, feel free to get whatever approval from stakeholders you need @ccleeton

ccleeton:
Discussed with PS. They use the request notes per transaction for staff use. Lets roll it out in the current model where it applies to every TN in the cart. (We might have to come back to this and have it apply to each transaction separately/possibly offer the option to apply to all or just one?) Can those options be investigated in the meantime?

The only other thing to roll this out in production would be to add a explanation/description to the note. Maybe a hover over bubble?

sdellis:
@ccleeton Hover bubbles don't work for mobile users, but maybe we can keep it short and keep it in the input field hint. What would the explanation/description say? Something like "Notes to Special Collections staff"?

ccleeton:
The sense I got was that we want users to know it will apply to all things in the cart. But I like the notes to staff part.

tpendragon:
Screen Shot 2021-03-29 at 12 52 15 PM

@ccleeton Is changing the placeholder text sufficient?

tpendragon:
I'm going to ship this, we can open another ticket if there's something we want to change about it.

ccleeton:
Sorry didn't see the question about the placeholder text before now. Looks good to me. I will see if there is more feedback after this has been in place for a while.

PR #774, opened 2021-03-23
Use child accessrestrict on component page. 

### LINKED ISSUE #773: Access restrictions in components are showing their parent's accessrestrict, even if they have one.
> Hi there, we just noticed something about how access restrictions are appearing in pulfalight. My understanding was that pulfalight would pull in the lowest level restriction note, all the way down to the file component level if there was one there. However, it looks like it pulls the lowest level series access restriction, and not the file level one. We confirmed that those file level access restriction notes were migrated to ASpace, so I don't think it's a matter of missing data.
Compare these components in pulfa2 and pulfalight. You can see that in pulfalight, it's displaying the access restriction for the series (which is a subseries nested several levels down), but in pulfa2, it's the file component one:
https://findingaids.princeton.edu/catalog/AC187_c00654
https://pulfa.princeton.edu/collections/AC187/c00654
Here is the component in ASpace (UA repo): https://aspace.princeton.edu/staff/resources/4185#tree::archival_object_1445850
We also looked at another collection and noticed the same thing.

Took a quick look, it looks like we're indexing the right thing in `accessrestrict_ssm` but we're displaying `parent_accessrestrict_ssm`.

tpendragon:
Closes #773


app/controllers/catalog_controller.rb
```diff
@@ -343,7 +343,7 @@ def item_requestable?(_ = nil, _options = {})
     config.add_component_field "language_ssm", label: "Languages"
     config.add_component_field "scopecontent_ssm", label: "Description", helper_method: :paragraph_separator
     config.add_component_field "bioghist_ssm", label: "Biography", helper_method: :paragraph_separator
-    config.add_component_field "parent_access_restrict_ssm", label: "Access Restrictions"
+    config.add_component_field "accessrestrict_ssm", label: "Access Restrictions"
```
jrgriffiniii, line 346: 👍

jrgriffiniii (approved):

PR #784, opened 2021-03-29
Delete unpublished records from Solr on sync.

### LINKED ISSUE #771: When a previously published collection is unpublished, it should be deleted from SOLR.
The `/ead/@audience` will be set to "internal."

Example is JTS001.

Deleting the parent via `delete_by_query("id:#{id}")` WILL NOT automatically delete the nested components. It might work to use the regular delete API, will have to test as part of this implementation.

tpendragon:
Closes #771


app/jobs/aspace_index_job.rb
```diff
@@ -58,4 +58,11 @@ def perform(resource_descriptions_uri:, repository_id: nil)
     blacklight_connection.add(solr_documents)
     logger.info("Successfully indexed the EAD")
   end
+
+  # Delete documents which are marked internal, in case they've been unpublished
+  # from ASpace, but were previously published.
+  def delete_document(xml_document)
+    ead_id = xml_document.xpath("//eadid").first.text
```
jrgriffiniii, line 65: 👍

jrgriffiniii (approved):

PR #796, opened 2021-04-05
XML Display for Collections

### LINKED ISSUE #107: Ensure that content negotiation exposes XML encoded finding aids
This is currently implemented in PULFA 2.0 in order to support the harvesting of collections and components by external organizations (please see https://findingaids.princeton.edu/collections/MC019/c01194 and https://findingaids.princeton.edu/collections/MC019/c01194.xml). A pattern is already structured for this in Blacklight with https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/routes/searchable.rb#L17, so this may be a minor enhancement.

We'll have to figure out how to save and manage these in a way that allows access to the EADs from multiple webservers.

## Implementation Proposal

* Create an AR Object with a text field, the call number string field, and an ID
* After downloading the XML stash it in postgres.
* For the pages just pull it from postgres if the record exists in SOLR.
* For component pages don't cache them, just parse it with Nokogiri and grab that component out.

tpendragon:
Closes #107

This adds an XML caching layer in ActiveRecord. It's the simplest low level cache store we have around and by utilizing one row per ead_id we don't have to worry about it getting too big or having a bunch of stale content.

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -17,7 +17,8 @@ class SolrDocument
   # single valued. See Blacklight::Document::SemanticFields#field_semantics
   # and Blacklight::Document::SemanticFields#to_semantic_values
   # Recommendation: Use field names from Dublin Core
-  use_extension(Blacklight::Document::DublinCore)
+  # use_extension(Blacklight::Document::DublinCore)
+  SolrDocument.use_extension(Pulfalight::Document::XMLExport)
 
   COLLECTION_LEVEL = "collection"
   ROOT_COMPONENT_LEVEL = 1
--- a/app/jobs/aspace_index_job.rb
+++ b/app/jobs/aspace_index_job.rb
@@ -43,7 +43,7 @@ class AspaceIndexJob < ApplicationJob
   end
 
   def perform(resource_descriptions_uri:, repository_id: nil)
-    ead_content = aspace_client.get("#{resource_descriptions_uri}.xml", query: { include_daos: true, include_unpublished: false }, timeout: 1200).body.force_encoding("UTF-8")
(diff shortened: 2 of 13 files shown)


app/models/solr_document.rb
```diff
@@ -17,7 +17,8 @@ class SolrDocument
   # single valued. See Blacklight::Document::SemanticFields#field_semantics
   # and Blacklight::Document::SemanticFields#to_semantic_values
   # Recommendation: Use field names from Dublin Core
-  use_extension(Blacklight::Document::DublinCore)
+  # use_extension(Blacklight::Document::DublinCore)
+  SolrDocument.use_extension(Pulfalight::Document::XMLExport)
```
jrgriffiniii, line 21: 👍

jrgriffiniii:
I'm assuming that there is no use case for cache expiry and refreshing the XML expression of the ArchivesSpace finding aid. Is this indeed the case?

tpendragon:
> I'm assuming that there is no use case for cache expiry and refreshing the XML expression of the ArchivesSpace finding aid. Is this indeed the case?

I don't think so. The indexing job skips the cache, so every time it's indexed it should re-cache the material.

faithc:
Is this only working for full the full finding aid and not individual pages/components?

tpendragon:
I think there's a bug for some components, but the component display should be working.

faithc:
Hmm, I've tried it in a couple of spots: https://findingaids.princeton.edu/catalog/C1491_c1.xml + https://findingaids.princeton.edu/catalog/C1491#description.xml.

tpendragon:
@faithc Created https://github.com/pulibrary/pulfalight/issues/799. Please prioritize appropriately.

PR #870, opened 2021-08-23
Standardize starting and stopping

bess:
Use rake servers:start and rake servers:stop, like we do in other
projects, to reduce team cognitive overhead.


tpendragon:
You're the best.

tpendragon (approved):

PR #882, opened 2021-08-27
Unescape HTML-escaped characters in SEARCHABLE_NOTES_FIELDS components

### LINKED ISSUE #806: Ampersands in scopecontent are escaped out of ASpace, but not unescaped at index time.
Screen Shot 2021-08-26 at 3 00 48 PM

See scope content note for https://findingaids.princeton.edu/catalog/C0140_c83445-31032 (Under "Description")

## Note

ASpace complains about ampersands, but it'll output &amp;. If we can't figure out the nice way to just decode all the special characters (because some may not be encoded), then just replace `&amp;` with &.

bess:
Fixes #806


spec/features/traject/ead2_indexing_spec.rb
```diff
@@ -416,6 +416,16 @@ def all_components(result)
         # Accruals - unable to find
       end
 
+      context "when names and descriptions contain ampersands" do
+        let(:fixture_path) do
+          Rails.root.join("spec", "fixtures", "aspace", "generated", "mss", "C0140.processed.EAD.xml")
+        end
+        it "escapes them before indexing" do
+          component = find_component(result, "C0140_c83445-31032")
+          expect(component["scopecontent_ssm"].first).to match(/J.P Ball & Son/)
```
jrgriffiniii, line 425: 👍

jrgriffiniii (approved):

PR #932, opened 2021-11-19
ItemInfo4 should choose FirstValue

### LINKED ISSUE #931: Location info not routing to Aeon/call slips
Top container profiles don't appear if multiple folders from a box are requested so staff won't know where the item is located. Not sure if this is a result of a character limit? In the example, "elephant size box" is missing. 
[containerprofilemissing.docx](https://github.com/pulibrary/pulfalight/files/7565610/containerprofilemissing.docx)
[containerprofilemissing2.docx](https://github.com/pulibrary/pulfalight/files/7565611/containerprofilemissing2.docx)

> This issue will result in items being very difficult to locate in the stacks.

Without this designation staff are forced to randomly browse stacks on the off chance material is there.

tpendragon:
It contains the box locator.

Fixes #931. Confirmed by @faithc

### DIFF
--- a/app/values/aeon_request.rb
+++ b/app/values/aeon_request.rb
@@ -183,6 +183,7 @@ class AeonRequest
       GroupingOption_ItemVolume: "FirstValue",
       GroupingOption_ItemInfo1: "FirstValue",
       GroupingOption_ItemInfo3: "Concatenate",
+      GroupingOption_ItemInfo4: "FirstValue",
       GroupingOption_Location: "FirstValue",
       SubmitButton: "Submit Request"
     }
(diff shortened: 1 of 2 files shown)


hackartisan (approved):
Would it be possible to refactor the test so the fields don't all have the same value? If so a follow up ticket for that would probably be good.

tpendragon:
> Would it be possible to refactor the test so the fields don't all have the same value? If so a follow up ticket for that would probably be good.

Huh?

hackartisan:
> > Would it be possible to refactor the test so the fields don't all have the same value? If so a follow up ticket for that would probably be good.
> 
> Huh?

I misread and thought "FirstValue" was an actual form value. I see now it's some sort of mapping type.

PR #963, opened 2022-01-18
Inherit accessrestrict from parents

### LINKED ISSUE #697: Address security vulnerability CVE-2015-9284 for omniauth
This follows a review and some discussion regarding the general practices around updating and patching security vulnerabilities for @pulibrary projects. After some very brief exploration, it appears that https://rubygems.org/gems/omniauth-rails_csrf_protection might not be the only solution now, and that https://rubygems.org/gems/omniauth/versions/2.0.0 offered a solution addressing the root cause of this vulnerability.

### LINKED ISSUE #924: Accessrestrict notes aren't inheriting from their parent, but from the root instead.
https://findingaids.princeton.edu/catalog/MC147_c07283-24964 should have pulled its accessrestrict note from its series, but it grabbed it from the collection instead. 

I think https://github.com/pulibrary/pulfalight/blob/fd732501daaa487df4bf0b1747d5bf41f214709a/lib/pulfalight/traject/ead2_component_config.rb#L572 should be `settings[:parent]` instead of using the clipboard, but it will need a test.

tpendragon:
This also fixes broken parent inheritance for several other components.

Closes #697


tpendragon:
This is ready for review - looking for word from @faithc that inheritance is working right.

tpendragon:
Phoebe confirmed that this looks good.

hackartisan (approved):

PR #1004, opened 2022-03-25
Don't mount the app root in Lando

tpendragon:
Shaves off ~ 18 seconds from the test suite, and these services don't need the app mount anyways.


hackartisan (approved):
cool!

Issue #899, opened 2021-09-08
Mismatching collection titles in aeon


ccleeton:
A recent student request had mismatching collection titles and call numbers. A student requested a box from C0714, but this request had the collection title of Carlos Baker Collection of Ernest Hemingway (C0365). This has happened once before with two different collections.
From Bri Cregle:
![TN 306750](https://user-images.githubusercontent.com/9676616/132560998-ec7490f7-fa8a-47de-93cb-7803124cb1b9.JPG)
![TN 306749](https://user-images.githubusercontent.com/9676616/132561002-60d3df12-083c-4a46-b926-15cde1562d38.JPG)

I did some testing and this happened to me, when I had a box from C0714 and C0365 in my shopping cart and requested both at the same time. Both requests from different collections had Carlos Baker Collection... as the collection title. This issue did not happen when I requested these two boxes separately.
Please let me know if you have any questions or need clarifications.

sdellis:
Test to see if we can send two Aeon requests with one button click.

hackartisan:
These screenshots are from the aeon client. The screen will look different when logged in as a test user looking at the request that you made - but you'll still be able to tell that it's wrong.

When making the request in pulfalight-staging, log in to aeon as the aeon test user (see lastpass). You should then be able to log in to the aeon "testweb" instance with that same user and see the bad request. Aeon admins will also see the request but they'll know not to fulfill it since it came from the test account.

tpendragon:
To replicate:

Request https://findingaids.princeton.edu/catalog/C0365_c0194, https://findingaids.princeton.edu/catalog/C0714_c025, and https://findingaids.princeton.edu/catalog/C0714_c026

It only happens if two items get combined within a box, and there's another box in another call number.

We need to look into the parameters that get sent in this case. We have a theory that it's falling back to the ItemTitle with no request number.

tpendragon:
FormData:

```
AeonForm: EADRequest
RequestType: Loan
DocumentType: Manuscript
Site: RBSC
Location: mss
ItemTitle: Ian Hamilton Working Papers for J. D. Salinger: A Writing Life
GroupingIdentifier: GroupingField
GroupingOption_ReferenceNumber: Concatenate
GroupingOption_ItemNumber: FirstValue
GroupingOption_ItemDate: FirstValue
GroupingOption_CallNumber: FirstValue
GroupingOption_ItemVolume: FirstValue
GroupingOption_ItemInfo1: FirstValue
[...]

PR #1016, opened 2022-04-18
Ensure the correct title shows up when grouped.

tpendragon:
Closes https://github.com/pulibrary/pulfalight/issues/899


docs/aeon_integration.md
```diff
@@ -0,0 +1,42 @@
...
+
+The form has a few different types of fields: global fields and per-request
+fields. Some of the global fields are Grouping Options.
+
+Grouping Options allows you to define a grouping field (ours is the call number
+and the box number) so that you can combine multiple requests into a single call
+slip. For each of the other values you can then define whether you want them
+concatenated or to just take the first value. Update [aeon_request.rb](https://github.com/pulibrary/pulfalight/blob/0cfda015d169ffdc34ddfeac81934382a147b307/app/values/aeon_request.rb#L174-L189) to change this logic.
+
+The other global fields we don't really because we submit multiple requests at
```
hackartisan, line 19: Turns out this isn't a sentence. Also I'm not sure it's entirely true. Maybe we should say something like "Otherwise we mostly prefer per-request fields over global fields because we submit multiple requests at once."

docs/aeon_integration.md
```diff
@@ -0,0 +1,42 @@
...
+once.
+
+### Implementation notes
+
+The view access the form data via a method on the solr_document. Each cart
```
hackartisan, line 24: accesses

hackartisan (changes requested):

hackartisan (approved):
🎉

PR #1040, opened 2022-04-27
Add oversize folders as top containers for requesting.

### LINKED ISSUE #755: Oversize folders should create a transaction in aeon per folder if there's multiple folders in a component
AC154_c03425 should create three transactions, one per folder.

Unfortunately ASpace isn't embedding the extra container information for those child containers to the EAD, so we'll have to update our plugin.

tpendragon:
Closes #755


hackartisan (approved):
🎉

PR #1043, opened 2022-04-28
Use a different field to pull id

### LINKED ISSUE #1042: internal only resources can't index because there's no id at the given path
It's looking for id here but unpublished EADs don't have it there. 

https://github.com/pulibrary/pulfalight/blob/b9e973b3b996fd5c8d47cc6843eb3223fadce535/lib/pulfalight/traject/ead2_config.rb

Maybe we can pull from `ead/archdesc/did/unitid`?

[PULFALight/staging] RSolr::Error::Http: RSolr::Error::Http - 400 Bad RequestError: {  "responseHeader":{    "status":400,    "QTime":3},  "error":{    "metadata":[      "error-class","org.apache.solr.common.SolrException",      "root-error-class","org.apache.solr.common.SolrException"],    "msg":"Document is missing mandatory uniqueKey field: id",    "code":400}}

## Backtrace

    line 57 of [PROJECT_ROOT]/app/jobs/aspace_index_job.rb: perform

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/65922/faults/85140485)

also resolve when this is fixed:
* https://app.honeybadger.io/projects/65922/faults/85157501 (prod issue)

hackartisan:
this one is populated even for unpublished eads

closes #1042

### DIFF
--- a/lib/pulfalight/traject/ead2_config.rb
+++ b/lib/pulfalight/traject/ead2_config.rb
@@ -36,13 +36,13 @@ configure_before
 
 # rubocop:disable Performance/StringReplacement
 # TODO: These should be combined into a single method
-to_field "id", extract_xpath("/ead/eadheader/eadid", to_text: false) do |_record, accumulator|
+to_field "id", extract_xpath("/ead/archdesc/did/unitid", to_text: false) do |_record, accumulator|
   string_array = accumulator.map(&:text)
   string_array = string_array.map { |a| a.gsub(".", "-") }
   string_array = string_array.map { |a| a.split("|").first }
   accumulator.replace(string_array)
 end
-to_field "ead_ssi", extract_xpath("/ead/eadheader/eadid", to_text: false) do |_record, accumulator|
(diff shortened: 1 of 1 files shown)


hackartisan:
Tested in staging and this does resolve the ticket and its linked honeybadger issues.

I'm still not sure about the fixtures for `C0879`. For the unpublished EADs I have looked at, all have the `audience` property in the top-level `ead` element. and nowhere else. The fixtures have that property all over the place. @regineheberlein has offered to work with aspace a bit more and see if there are any records that are unpublished at the component or even field level (based on what is in those fixtures).

hackartisan:
@faithc that's helpful, thank you. If you happen to know an example that would be very helpful.

Do you know if there's a concept of "unpublished" or "internal access" on the field-level?

tpendragon (changes requested):
We had some conversation in Slack. I'm just blocking this so we can test it on staging first and add some documentation about how the generation doesn't include unpublished content.

In general, great work!

tpendragon:
Staging's been locked up trying to index with this branch for days - we'll have to figure that out before we can get this in, or come up with a different strategy.

PR #1108, opened 2022-07-29
Increase sticky bar contrast

### LINKED ISSUE #1012: Sticky bar needs more contrast
Making the sticky bar of action items (request this box, etc.) a bit darker or more apparent

taylor-yamashita:
closes #1012 

- [x] expand sticky bar horizontally to edges
- [x] change sticky bar and top nav bar background color (reference other blues on page/lux)
- [x] change sticky bar outline color to be more consistent 
- [x] change sticky bar text/links color (matching top nav bar)


hackartisan (changes requested):
This definitely achieves what the ticket requested, but looking at it I think we might want to do some more design work here. The darker background accentuates the fact that this area doesn't fill the space horizontally. I also wonder if we need more vertical breathing room between the content list (or item details) and this element. And when it's the same color as the footer does it need to be distinguished from the footer in some way?

sdellis:
@hackartisan or @taylor-yamashita Can you upload a screenshot for reference?

taylor-yamashita:
Before changes:
before

With changes:
after

taylor-yamashita:
@sdellis here are the screenshots:
Before changes:
before
With changes:
after

taylor-yamashita:
image

taylor-yamashita:
@hackartisan I made some more changes after consulting with Shaun!

Sticky bar fills width of div: 
image

Sticky bar has more contrast with footer:
image

Sticky bar link colors/borders styled to be consistent with top navbar:
image

app/assets/stylesheets/_overrides.scss
```diff
@@ -44,9 +44,9 @@ a {
 }
 
 .lux-tabs-item {
-  background: #f8f9fa;
+  background: #e4e4e4;
```
hackartisan, line 47: For the places in this PR where we have hex values, I'd love to instead have named variables in https://github.com/pulibrary/pulfalight/blob/main/app/assets/stylesheets/variables/colors.scss.

app/views/shared/_header_navbar.html.erb
```diff
@@ -1,5 +1,5 @@
 <header class="lux">
-  <banner>
+  <banner dismissible>
```
hackartisan, line 2: I think it would be good to have this change in a separate PR

hackartisan (changes requested):
@taylor-yamashita it looks really nice!! Please see a couple of inline comments.

It's resulting in a test failure about contrast required to meet accessibility standards. @m-ichelleliu saw this same issue when working on https://github.com/pulibrary/pulfalight/pull/1110.

@sdellis do you think we should consider changing the link color for the site?

app/assets/stylesheets/_overrides.scss
taylor-yamashita: Addressed this.

app/views/shared/_header_navbar.html.erb
taylor-yamashita: Sounds good, will do!

sdellis: Sorry, @taylor-yamashita ! That was my fault. You asked me, and I thought it was a small enough change that you could add it here, but I was wrong.

hackartisan:
Hm, there are contrast failures here still?

app/assets/stylesheets/_overrides.scss
sdellis: @hackartisan I don't think we need to change the site link color. I'm a bit confused how the tabs have been passing accessibility tests but the sticky bar doesn't if they're consistent. But regardless, @taylor-yamashita should feel free to experiment with bg colors and link text colors to make this look good and pass contrast tests. (We can also make the link text white.) I see these more as buttons (or at least look more like buttons and less like links) and would love to see dividers between them. Same with the tabs.

taylor-yamashita: For now, I will experiment with the colors. If we add dividers between the buttons/links/tabs, maybe that can go in a separate ticket?

taylor-yamashita: I ended up using `$bleu-de-france-darker: #23578b;` (a slightly darker blue pulled from the lux palettes) for the links.

app/assets/stylesheets/_overrides.scss
```diff
@@ -61,3 +61,8 @@ a {
 .lux-tabs-container .active {
   background: white;
 }
+
+.lux-button.text, .lux-tabs-container a {
+  color: $bleu-de-france-darker;
```
taylor-yamashita, line 66: This is the change that handled the failing Circle CI contrast test.

hackartisan (approved):
👍🏻 

Screen Shot 2022-09-21 at 9 18 41 AM

Issue #1045, opened 2022-04-29
Can the "Welcome to the new Finding Aids site!" banner come down?


sdellis:
I understand that the general public does not have the same access policy as the Princeton community due to the pandemic, so there may be a desire to continue to inform the public about access restrictions/options for the public. If this is going to be a long-term policy, I think it would be best to integrate it into the site as the banner is really intended for information that is temporary in nature. 

@ccleeton 
@slogue

slogue:
Shaun, I'm hopeful that the COVID policies are temporary and will be lifted at some point in the near future. Until that happens, though, I would like to keep the banner in place. Thanks!

ccleeton:
I would at least like to take the New Finding Aids text down and just leave the access link

ccleeton:
@sdellis I was able to confirm with @slogue today that we can remove the New finding aids text and just leave the access link.

Is this something that we can do out of cycle?

ccleeton:
@slogue Would you be ok with moving the covid info out of the banner

slogue:
At this point, most of the COVID restrictions have lifted, so I think we can just take it out of the banner. I was waiting for that to happen and then forgot about this ticket.

ccleeton:
@sdellis So we can get rid of all the top banner!!!

PR #1118, opened 2022-09-06
Make COVID announcement banner dismissible

taylor-yamashita:
This was a minor change that came up when working with @sdellis last month. I added the "dismissible" keyword to the banner element but don't know how to make the color of the actual "x" darker for more contrast because I lack lux/vue knowledge...

### DIFF
--- a/app/views/shared/_header_navbar.html.erb
+++ b/app/views/shared/_header_navbar.html.erb
@@ -1,5 +1,5 @@
 <header class="lux">
-  <banner>
+  <banner dismissible>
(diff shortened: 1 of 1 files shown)


hackartisan:
Hm, you're right it's hard to see. Maybe we should wait on this until we can figure out how to make it darker. Here's a screenshot.

Screen Shot 2022-09-07 at 9 21 46 AM

ccleeton:
There is also the ticket #1045 which asks to take down the Welcome to the new finding aids text. Which would make the banner smaller.

hackartisan:
We were advised that the banner could be removed; done in #1124

PR #1124, opened 2022-09-09
Delete the Welcome/COVID information banner

### LINKED ISSUE #1045: Can the "Welcome to the new Finding Aids site!" banner come down?

m-ichelleliu:
Resolves #1045


hackartisan (approved):
Looks great!

PR #1130, opened 2022-09-15
Update Pulfalight for Ruby 3.1

### LINKED ISSUE #1125: Update ruby to 3.0
2.7 EOL is march 2023. 3.0 EOL is March 2024.

If 3.1 is straightforward, go ahead and do it.

tpendragon:
Closes #1125

Draft until we actually do it in Ansible.


app/overrides/arclight/helper_overrides.rb
```diff
@@ -1,12 +0,0 @@
-# frozen_string_literal: true
-
-# Overrides to Arclight helper methods
-module HelperOverrides
```
tpendragon, line 4: I have no clue why this isn't necessary anymore, but I tested the functionality and it still works, and also the test is still passing.

hackartisan (approved):

Issue #817, opened 2021-05-10
Labels for description notes should display for collection records


faithc:
Data/headings that are added to the Label field in multipart notes are not displaying in pulfalight. For example, the Processing Info Conservation note in this mss repo resource record: https://aspace.princeton.edu/staff/resources/4187/edit#tree::resource_4187 does not display as such: https://findingaids.princeton.edu/catalog/C1664#collection-history. Generally, label headings in ASpace should export; rather than the default heading. 

## Technical Implementation

We store the headings in a `processinfo_heading_ssim` etc. Prefer those headings. For instance, the second processinfo note in this example should have the heading of "Conservation"

## Example of Collection Record

https://findingaids.princeton.edu/catalog/TC040#description

This should have a Content Warning header.

tpendragon:
It seems like people who do this are defining custom ViewComponents, but I don't think Arclight uses that Blacklight release yet.

ccleeton:
@tpendragon Can you let us know if this is still blocked? If so we will take it out of the list.

tpendragon:
I'd consider it still blocked, yes.

faithc:
Dang it. Any idea when ArcLight might adopt the Blacklight release that allows for this? (If I'm understanding your previous comment.)

tpendragon:
Oh, it looks like maybe they do have it now - we'd have to update Arclight. If this is a high priority you can keep it on deck, but be forewarned that it's going to be hard.

tpendragon:
The example given still doesn't have the right headings, even though it worked in staging and some others are working.

tpendragon:
Oh, yes it is, I must have had it cached.

faithc:
Out of curiosity, could this functionality be applied to Dates as well as notes fields? We typically use the "creation" label, but may also use others as appropriate from ASpace's drop-down menu as appropriate. Would I need to create another ticket for this? Thanks.

hackartisan:
@faithc yes a new ticket please thank you!

faithc:
This functionality is not being applied at the collection level as it should be. Please see for example, the [description note here](https://findingaids.princeton.edu/catalog/TC040#description) does not include the "Content Warning" label even though it's employed in ASpace: https://aspace.princeton.edu/staff/resources/2543#resource_notes_  (mss repo)

tpendragon:
[...]

PR #1134, opened 2022-09-15
Display EAD headings alongside metadata. 

tpendragon:
This is achieved by indexing a JSON combination of the headers with
their fields, then displaying those.

We should be able to deploy this and run indexing, as before the
combined field exists it falls back to the previous implementation.

Closes https://github.com/pulibrary/pulfalight/issues/817


tpendragon:
I confirmed in staging that the fallback works correctly.

eliotjordan (approved):

PR #1137, opened 2022-09-16
Stick footer to the bottom of the page

### LINKED ISSUE #1123: stick footer to bottom of page
It floats up on the home page, leaving a large white space below it.

screenshot from Brave:
Screen Shot 2022-09-09 at 11 58 26 AM

taylor-yamashita:
Closes #1123 

Modeled #main-container styling after the one in DPUL to make footer sticky.

### DIFF
--- a/app/assets/stylesheets/_overrides.scss
+++ b/app/assets/stylesheets/_overrides.scss
@@ -14,7 +14,7 @@ $container-max-widths: (
 ) !default;
 
 #main-container {
-  min-height: calc(100vh - 300px);
+  min-height: calc(100vh - 600px);
   margin-bottom: 20px;
 }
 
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/_overrides.scss
```diff
@@ -13,6 +13,11 @@ $container-max-widths: (
   xl: 1400px
 ) !default;
 
+#main-container {
+  min-height: calc(100vh - 300px);
```
thanyabegum, line 17: Could we subtract more than 300px? Right now you have to scroll to see the footer on bigger screens, but it should be completely visible initially.

thanyabegum (changes requested):

app/assets/stylesheets/_overrides.scss
taylor-yamashita: Yes! Fixed

thanyabegum (approved):

PR #1141, opened 2022-09-19
Add new physloc codes

### LINKED ISSUE #1136: add new physloc codes
We're adding 9 new location codes to finding aids; could you add them to pulfalight please? 

Background: the reason we're doing this is so we can tell apart material that's requestable from the catalog vs. material that is in the catalog but should be requested from the finding aids; we'll use the physloc codes to suppress the "request in reading room" button in orangelight. 

The display values for the new locations should be the same as their equivalents as follows:

```
 scarcpxm (same as rcpxm)
 scactsn (same as ctsn)
 scagax (same as gax)
 scamss (same as mss)
 scawa (same as wa)
 scaex (same as ex)
 scahsvm (same as hsvm)
 scamudd (same as mudd)
 scathx (same as thx)
```

See related https://github.com/pulibrary/orangelight/issues/2910

eliotjordan:
Closes #1136


config/physical_location_codes.yml
```diff
@@ -9,3 +9,12 @@ publicpolicy: 'MUDD'
...
 rcpph: 'MUDD'
+scarcpxm: 'Remote Storage (ReCAP): Manuscripts. Special Collections Use Only'
+scactsn: 'Cotsen Children’s Library Archival. Special Collections Use Only'
+scagax: 'Graphic Arts Archival. Special Collections Use Only'
+scamss: 'Manuscripts Archival. Special Collections Use Only'
+scawa: 'Western Americana Archival. Special Collections Use Only'
+scaex: 'Rare Books Archival. Special Collections Use Only'
+scahsvm: 'Manuscripts Archival. Special Collections Use Only'
+scamudd: 'Mudd Archival. Special Collections Use Only'
+scathx: 'Theater Archival. Special Collections Use Only'
```
eliotjordan, line 20: @regineheberlein Are these values correct? I wasn't sure if I should use Internal or External names from the Location Proposal Checklist document.

regineheberlein: Thanks @eliotjordan! The archivists are applying the internal name (i.e. code) to the description, and we'll want to expand that for display on the finding aids site and also make sure the code gets passed on to Aeon for requesting. 

I'm not entirely sure if I'm answering your question, let me know if not!

eliotjordan: According to this [Google Doc](https://docs.google.com/document/d/1E36NShlpqeq7zFXmMcP1Tu5ZD2NWgySgp20teTGiQWE/edit#), there is an Internal Name for staff:
 `scaex: 'Rare Books Archival'`

and an External Name for public display:
`scaex: 'Rare Books Archival. Special Collections Use Only'` 

I chose the external name. It sounds like we should use the internal name instead?

sdellis (approved):
👍

regineheberlein:
@eliotjordan I'm sorry to come back to this--it looks like I need to reverse myself. We need to *display* the external name in pulfalight, but we need to *pass to Aeon* the internal name (i.e. the code) for requesting. It looks like right now, what's getting passed to Aeon is the external name. Could you fix that? Thanks, and sorry again!

PR #1157, opened 2022-09-23
Remove audience filters

### LINKED ISSUE #1010: Figgy should synchronize finding aid data that hasn't yet been published in aspace
proxy for pulibrary/figgy#5050

hackartisan:
Once prod is finished reindexing these won't be needed anymore because
there won't be any unpublished records in the index.

closes #1010


hackartisan:
Checked on staging -- full results counts are the same before / after deploying this PR. 👍🏻

PR #1176, opened 2022-10-03
Allow collapsing of Table of Contents

### LINKED ISSUE #111: Ensure that mobile devices (smartphones and tablets) are able to view and interact with PULFALight
This support is limited at best for smartphones:

image

image

thanyabegum:
Closes #111 by creating a collapsible Table of Contents so scrolling on mobile devices isn't trapped within it. Now, the Table of Contents will be collapsed by default on mobile devices.

Before
Screen Shot 2022-10-03 at 4 40 59 PM

After (Expanded)
Screen Shot 2022-10-10 at 2 21 28 PM

After (Collapsed)
Screen Shot 2022-10-10 at 2 21 16 PM

### DIFF
--- a/app/javascript/pulfalight/pulfalight_loader.es6
+++ b/app/javascript/pulfalight/pulfalight_loader.es6
@@ -13,6 +13,7 @@ import LibCalHours from "../pulfalight/lib_cal_hours.es6"
 import ChildTable from "../components/ChildTable"
 import PulfaDataTable from "../components/PulfaDataTable"
 import FiggyViewer from "../components/FiggyViewer"
+import MediaQueries from "../pulfalight/media_queries.es6"
 
 export default class {
   run() {
@@ -21,7 +22,7 @@ export default class {
     this.setup_lib_cal_hours()
     this.setup_range_limit()
     this.setup_form_modal()
-    this.setup_toc_collapse()
+    this.setup_media_queries()
   }
 
   setup_lib_cal_hours() {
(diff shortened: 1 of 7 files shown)


app/javascript/pulfalight/pulfalight_loader.es6
```diff
@@ -45,6 +46,17 @@ export default class {
     toc.build()
   }
 
+  setup_toc_collapse() {
+    const mediaQuery = window.matchMedia('(max-width: 768px)')
```
hackartisan, line 50: Where does 768 come from? If we can't use a variable defined somewhere else, let's add a comment pointing to where it comes from.

app/javascript/pulfalight/pulfalight_loader.es6
```diff
@@ -45,6 +46,17 @@ export default class {
...
+    function handle_mobile_change(e) {
```
hackartisan, line 51: I'm not sure how I feel about having this function defined inside the other function. Not a good sense of javascript style conventions here, but my initial reaction is that I'd like to see it defined outside.

hackartisan (changes requested):
Maybe we could chat about all my javscript questions? 😂 

I spent too much time trying to figure out best practices for which way the carat should point. I mostly end up where you are, but I also notice that the facets on the search results page are different. https://findingaids.princeton.edu/?utf8=%E2%9C%93&group=true&search_field=all_fields&q= I wonder if we should make a ticket to bring consistency between these two elements.

app/javascript/pulfalight/pulfalight_loader.es6
thanyabegum: I'll add a comment referencing the file containing the breakpoints for the application. I'll make a ticket since this seems like an inconsistency throughout the repo.

app/javascript/pulfalight/pulfalight_loader.es6
```diff
@@ -45,6 +46,17 @@ export default class {
...
+      if (e.matches) {
```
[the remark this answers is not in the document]
thanyabegum: `e` is the media query event and `matches` is a read only property of it (see [here](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList)), so I'll add a comment clarifying that.

hackartisan (approved):
looks great and works great! 🎉

PR #1177, opened 2022-10-04
Display "Other Finding Aids" under "Find Related Materials"

### LINKED ISSUE #967: Other Finding Aids note should be display under Find Related Materials
The "Other Finding Aids" note should be associated with/display under the Find Related Materials tab rather than the Access & Use tab where is currently displays.

taylor-yamashita:
closes #967


hackartisan (approved):
👍🏻 thank you!

PR #1184, opened 2022-10-11
Make search bar autocomplete text invisible

### LINKED ISSUE #1171: Make autocomplete text invisible in search bar
Set the opacity to 0. See this blacklight commit for guidance: https://github.com/projectblacklight/blacklight/pull/2807/files

child of #1165

taylor-yamashita:
closes #1171 

I decided to set the text color to white instead of using the suggested `filter: opacity(0);` because doing the latter made the whole search bar invisible.

### DIFF
--- a/app/assets/stylesheets/modules/search.scss
+++ b/app/assets/stylesheets/modules/search.scss
@@ -118,7 +118,7 @@ header .navbar-form .input-group .input-group-btn {
 }
 
 .search-q.form-control.tt-hint {
-  color: white;
+  visibility: hidden;
 }
 
 @media (min-width: $bp-large) {
--- a/app/assets/stylesheets/modules/search.scss
+++ b/app/assets/stylesheets/modules/search.scss
@@ -117,6 +117,10 @@ header .navbar-form .input-group .input-group-btn {
   height: $search-box-form-height;
 }
 
+.search-q.form-control.tt-hint {
+  color: white;
+}
+
 @media (min-width: $bp-large) {
   .navbar-form .input-group >
   .search-q.form-control {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/modules/search.scss
```diff
@@ -117,6 +117,10 @@ header .navbar-form .input-group .input-group-btn {
   height: $search-box-form-height;
 }
 
+.search-q.form-control.tt-hint {
+  color: white;
```
hackartisan, line 121: Can we try setting it to zero opacity, similar to the way it's done in https://github.com/projectblacklight/blacklight/pull/2807/files

hackartisan (changes requested):

app/assets/stylesheets/modules/search.scss
taylor-yamashita: Hello! Please see my initial comment above. When I set 0 opacity, it looks like this: 
image

I tried a number of different selectors but wasn't able to isolate just the autocomplete text--any suggestions?

hackartisan: Ah! sorry I  missed that.

app/assets/stylesheets/modules/search.scss
```diff
@@ -118,7 +118,7 @@ header .navbar-form .input-group .input-group-btn {
 }
 
 .search-q.form-control.tt-hint {
-  color: white;
+  visibility: hidden;
```
hackartisan, line 121: This definitely works to make it invisible but from looking at the selectors it's making the input element itself invisible (which I guess is why the white background needed to move to the parent element). I think this will be an accessibility issue in that now screen readers will skip the search input.

I found a new idea that I will put in the ticket.

hackartisan (changes requested):

PR #1191, opened 2022-10-14
Display consecutive containers as a range

### LINKED ISSUE #901: Display ranges for consecutively-numbered containers with AbIDs
Top containers with AbIDs are currently displaying separately rather than as ranges when there are consecutive numbers. See for example https://findingaids.princeton.edu/catalog/C1643.  

Related to #897 (storage notes)

m-ichelleliu:
Resolves #901


m-ichelleliu:
Will try to pair with Thanya on Monday to clean up the code / address assumptions

app/models/solr_document.rb
```diff
@@ -150,9 +150,35 @@ def extents
   end
 
   def fetch_html_safe(field)
+    return fetch_summary_note(field) if field == "summary_storage_note_ssm"
     fetch(field, []).map(&:html_safe)
   end
 
+  def fetch_summary_note(field)
+    note = fetch(field, [])
+    return note.map(&:html_safe) if note == [] || !note[0].match(/Firestone Library \(mss\): Boxes (?:[A-Z]-\d{6}; )+/)
```
hackartisan, line 159: We talked about using a named capture group more like `/^(.*: Boxes )(?:[A-Z]-\d{6}; )+/` so we don't have to hard code the library name.

app/models/solr_document.rb
```diff
@@ -150,9 +150,35 @@ def extents
...
+    location = "Firestone Library (mss): Boxes "
+    boxes = note[0].scan(/[A-Z]-\d{6}/).sort
```
hackartisan, line 161: `note.first` is slightly more idiomatic

app/models/solr_document.rb
hackartisan: For longer postfix conditionals like this it's nice to put them in a small private method and give them a name for readability.

app/models/solr_document.rb
```diff
@@ -150,9 +150,35 @@ def extents
...
+  end
+
+  def consecutive?(box1, box2)
+    return false if box1.gsub(/\d/, "") != box2.gsub(/\d/, "") # compare starting letter
+    box2.gsub(/\D/, "").to_i == box1.gsub(/\D/, "").to_i + 1 # if same letter, then compare the integer
+  end
+
+  def consecutive_chunk(output, boxes)
+    first = boxes[0]
+    boxes.shift while (boxes.length > 1) && consecutive?(boxes[0], boxes[1])
```
hackartisan, line 172: Let's try turning this back into a more traditional while loop structure.

hackartisan (changes requested):
This is looking really good and much more readable

app/models/solr_document.rb
```diff
@@ -150,9 +150,37 @@ def extents
   end
 
   def fetch_html_safe(field)
+    return fetch_summary_note(field) if field == "summary_storage_note_ssm"
     fetch(field, []).map(&:html_safe)
   end
 
+  def fetch_summary_note(field)
+    note = fetch(field, [])
+    matches_abid_format = note.first.match(/^(?<location>.*: Boxes )(?:[A-Z]-\d{6}; )+/)
```
hackartisan, line 159: I think this will error if note is an empty array. Then `note.first` will be `nil` and `match` will be undefined. It would probably be good to add a test for this case so we can see it fail and then fix that bug.

I think it will show that we need to check for that nil before we run the match. What about a refactor like

```
note = fetch(field, [])
if note.present?
  abid_matcher = note.first.match(...)
  if abid_matcher
    boxes = note.first.scan(/[A-Z]-\d{6}/).sort
    note = ["#{abid_matcher[:location]}#{boxes_to_range(boxes)}"]
  end
end note.map(&:html_safe)
```

This also has the benefits of expanding the postfix conditionals, and only running the map once.

app/models/solr_document.rb
```diff
@@ -150,9 +150,37 @@ def extents
...
+    return note.map(&:html_safe) if note == [] || !matches_abid_format
+    boxes = note.first.scan(/[A-Z]-\d{6}/).sort
+    ["#{matches_abid_format[:location]}#{boxes_to_range(boxes)}"].map(&:html_safe)
+  end
+
+  def consecutive?(box1, box2)
+    same_prefix = box1.gsub(/\d/, "") == box2.gsub(/\d/, "")
+    return false unless same_prefix
+    box_numbers_consecutive = (box2.gsub(/\D/, "").to_i == box1.gsub(/\D/, "").to_i + 1)
```
hackartisan, line 168: I don't think we need this variable; the comment was enough here.

app/models/solr_document.rb
```diff
@@ -150,9 +150,37 @@ def extents
...
+    note = fetch(field, [])
+    matches_abid_format = note.first.match(/^(?<location>.*: Boxes )(?:[A-Z]-\d{6}; )+/)
+    return note.map(&:html_safe) if note == [] || !matches_abid_format
+    boxes = note.first.scan(/[A-Z]-\d{6}/).sort
+    ["#{matches_abid_format[:location]}#{boxes_to_range(boxes)}"].map(&:html_safe)
+  end
+
+  def consecutive?(box1, box2)
+    same_prefix = box1.gsub(/\d/, "") == box2.gsub(/\d/, "")
+    return false unless same_prefix
```
hackartisan, line 167: this is easier to read thank you!

app/models/solr_document.rb
```diff
@@ -150,9 +150,37 @@ def extents
...
+    box_numbers_consecutive = (box2.gsub(/\D/, "").to_i == box1.gsub(/\D/, "").to_i + 1)
+    box_numbers_consecutive
+  end
+
+  def consecutive_chunk(output, boxes)
+    first = boxes.first
+    boxes.shift while boxes.length > 1 && consecutive?(boxes[0], boxes[1])
```
hackartisan, line 174: `boxes.present?` might be a bit more readable than `boxes.length > 1` since it doesn't have any spaces.

app/models/solr_document.rb
```diff
@@ -150,9 +150,37 @@ def extents
...
+    return output.push(boxes.shift), boxes if first == boxes.first
```
hackartisan, line 175: can you add the square brackets around the array / tuple here to make it more obvious like in the line below?

app/models/solr_document.rb
```diff
@@ -150,9 +150,37 @@ def extents
...
 
+  def fetch_summary_note(field)
+    note = fetch(field, [])
+    matches_abid_format = note.first.match(/^(?<location>.*: Boxes )(?:[A-Z]-\d{6}; )+/)
+    return note.map(&:html_safe) if note == [] || !matches_abid_format
+    boxes = note.first.scan(/[A-Z]-\d{6}/).sort
+    ["#{matches_abid_format[:location]}#{boxes_to_range(boxes)}"].map(&:html_safe)
+  end
+
+  def consecutive?(box1, box2)
```
hackartisan, line 165: I think we should move this method and the ones below it to be private.

hackartisan (changes requested):

hackartisan (approved):

PR #1253, opened 2023-07-03
Update content icon in TOC to be grayed out if not direct

### LINKED ISSUE #1160: Change TOC icon for branch-level (series) nodes that may have some online components
These icons would be the same as the green ones, but instead they would have the following HEX color: #636363

Example that should have this hex: https://findingaids.princeton.edu/catalog/C0022_c0089

tpendragon:
Closes #1160 

Screen Shot 2023-07-03 at 12 46 51 PM


tpendragon:
Here's a shot at lighter green:

Screen Shot 2023-07-03 at 12 49 52 PM

tpendragon:
@eliotjordan @faithc Can you two look at these two color options and see what looks good?

eliotjordan:
@tpendragon The lighter green has much more contrast. I have a really hard time distinguishing the difference on the first example.

hackartisan:
> @tpendragon The lighter green has much more contrast. I have a really hard time distinguishing the difference in the first example.

@eliotjordan contrast from the darker green or contrast from its background?

eliotjordan:
From the darker green. I.e. I can tell that the icons are different colors with the light green variant.

faithc:
Actually, I think I prefer the grey. Now that I’m seeing it, I feel like the two different color greens might be more confusing. 

@ccleeton What say you?

tpendragon:
> From the darker green. I.e. I can tell that the icons are different colors with the light green variant.

Maybe there's another color? The bright green really draws my eyes in a way that I'm not sure we want it to. Realistically we probably need a new icon, but I'm not sure we'll be able to fit that into the cycle. Thoughts @sdellis ?

sdellis:
I tried messing with some icons that don't use colors to differentiate between direct and indirect online content. I'm not happy with the outcome, but here they are so everyone can review.  

has_online_content_24x24
some_online_content-24x24
some_online_content
has_online_content

Also, it turns out that the green we use in the TOC does not pass the WCAG AA Accessibility guidelines for contrast. Green Hex Number #397339 is the lightest green that is AA acceptable at that text size with the current TOC background colors. In order to get the "breadcrumbing" effect we want, we should change the background of those that have direct online content to match the icon color. Here is a mockup of what that would look like using Green #397339: 

pulfalight-accessible

tpendragon:
https://github.com/pulibrary/pulfalight/tree/grayed-out-icons-w-bar has some CSS that should make some of these suggestions easier to implement.

pulfalight-online-access-mockup-variant1

is what that looks like right now.

sdellis:
Here are some mockups I created from our DLS stand-up discussion yesterday. Thoughts? @faithc what do you think? (click the image to see an enlarged version)

Option A:
pulfalight-online-access-mockup-bars

Option B:
pulfalight-online-access-mockup-folders

Option C:
pulfalight-online-access-mockup-plus-minus

Option D:
pulfalight-online-access-mockup

tpendragon:
Added commit for the green bars. It looks like this now:

sdellis:
Screenshot

eliotjordan (approved):

PR #1257, opened 2023-07-05
Add table of contents online toggle.

### LINKED ISSUE #1162: Create a toggle to filter show only online content (and series that have online content) from the collection page view
The toggle allows browsing, with hierarchy, of only online content in a collection. It should instantly filter them out, without requiring a page reload. 

Screen Shot 2022-09-26 at 2 47 52 PM

## Implementation Idea

Adjust https://github.com/pulibrary/pulfalight/blob/5bb71c5d8b13a0cf420b986bb93c04f46c0e0eaa/app/services/table_of_contents_builder.rb#L42-L49

so that it adds `li_attr: { class: "online-content" }` (documented here: https://www.jstree.com/docs/json/) if the level or any of its children has online content (should just be able to check has_online_content_ssim), and another class if it doesn't,

Add a style that says if the toc node has the class "online-filter" then any li with the no-online-content class gets `display: none`, then make a toggle that toggles that class on the root node.

tpendragon:
Closes #1162

Screen Shot 2023-07-05 at 10 26 18 AM
Screen Shot 2023-07-05 at 10 26 22 AM

### DIFF
--- a/app/assets/stylesheets/modules/toc.scss
+++ b/app/assets/stylesheets/modules/toc.scss
@@ -1,3 +1,10 @@
+#toc-container {
+  &.online-only {
+    li[data-online-content="false"] {
+      display: none;
+    }
+  }
+}
 #toc {
   margin-bottom: 20px;
   max-height: 800px;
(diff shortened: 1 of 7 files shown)


eliotjordan (approved):
Nice job using css to hide non-online items. I wonder if the toggle button is too small? On the other hand, it doesn't take up much room on the page.

PR #1260, opened 2023-07-05
Enhance online content tooltip

### LINKED ISSUE #1163: Update the icon tooltip in the TOC to have clearer info
The tooltip should say "has online content" or "some online content" instead of simply "online"

eliotjordan:
Closes #1163 

Screenshot 2023-07-13 at 12 39 31 PM

Screenshot 2023-07-13 at 12 54 54 PM

### DIFF
--- a/app/assets/stylesheets/modules/_document_access_badge.scss
+++ b/app/assets/stylesheets/modules/_document_access_badge.scss
@@ -1,6 +1,12 @@
+.online-direct-content {
+  & svg {
+    pointer-events: none;
+  }
+}
 .online-indirect-content {
   & svg {
     fill: #636363;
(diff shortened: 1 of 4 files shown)


sdellis:
Thoughts on these icons to represent Direct and Indirect Online Content to avoid differences in color? We could also use a dotted line for the window view icon if that's the preferred icon. My only hesitation is that some content is audio-only so the eye may not be the best fit. We could attempt a similar dashed circle/full circle adaption of the current icon. (These are transparent icons, so dark mode on GitHub may make them hard to see here.)

dashed-circle-view
full-circle-view
window-view 1

tpendragon (approved):

PR #1268, opened 2023-07-10
Fix JS loading issues with TOC

eliotjordan:
- Build TOC only when element exists in document
- Scroll only when a element is selected

Confirmed that this is working in staging.

### DIFF
--- a/app/javascript/pulfalight/toc.es6
+++ b/app/javascript/pulfalight/toc.es6
@@ -30,8 +30,10 @@ export default class TocBuilder {
   }
 
   build() {
-    this.setupTree()
-    this.setupEventHandlers()
+    if (this.element.length > 0) {
+      this.setupTree()
+      this.setupEventHandlers()
+    }
   }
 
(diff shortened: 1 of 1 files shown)


app/javascript/pulfalight/toc.es6
```diff
@@ -43,8 +45,10 @@ export default class TocBuilder {
     this.element.on('ready.jstree', (e, data) => {
       const selectedId = this.element.jstree().get_selected()[0]
       const selectedElement = $(`#${selectedId}`)
-      const scrollOffset = selectedElement.offset().top - selectedElement.offsetParent().offset().top - 60
-      this.element.scrollTop(scrollOffset)
+      if (selectedElement.length > 0) {
```
tpendragon, line 48: Suggestion: Do you know what this is for exactly? I was pretty confused by it, a comment might help if you know what's going on?

tpendragon (approved):

tpendragon (approved):

PR #1343, opened 2023-11-15
Add functionality to switch search algorithms

### LINKED ISSUE #1342: Add search algorithm switching UI from Orangelight
See pulibrary/orangelight#3744

This will enable us to try new algorithms, like ones which will solve #1341.

## Implementation Details

We don't have FlipFlop, so we might have to use some sort of config to enable/disable this?

- [ ] update solr configs as in https://github.com/pulibrary/pul_solr/pull/375 and https://github.com/pulibrary/pul_solr/pull/377
- [ ] add backend https://github.com/pulibrary/orangelight/pull/3819
- [ ] add UI https://github.com/pulibrary/orangelight/pull/3827

As an example, boost results with the online facet.

eliotjordan:
Closes #1342 

Screenshot 2023-11-15 at 11 35 20 AM

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -119,6 +119,7 @@ class CatalogController < ApplicationController
     # config.show.thumbnail_field = 'thumbnail_path_ss'
 
     config.add_results_collection_tool(:sort_widget)
+    config.add_results_collection_tool(:algorithm_select)
     config.add_results_collection_tool(:per_page_widget)
     config.add_results_collection_tool(:view_type_group)
 
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -5,6 +5,7 @@ class CatalogController < ApplicationController
   include Arclight::Catalog
   include Arclight::FieldConfigHelpers
   include BlacklightRangeLimit::ControllerOverride
+  include MultipleAlgorithms
 
   # This should be unnecessary with an update ArcLight or Blacklight upstream
   def raw
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -70,6 +70,10 @@ class CatalogController < ApplicationController
     super
   end
 
+  # Set allowed search alorithm parameter values
+  # See: multiple_algorithms.rb
+  MultipleAlgorithms.allowed_search_algorithms = ["default", "online_content"]
+
   configure_blacklight do |config|
     ## Class for sending and receiving requests from a search index
     # config.repository_class = Blacklight::Solr::Repository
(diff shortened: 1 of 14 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -118,6 +119,7 @@ def index
     # config.show.thumbnail_field = 'thumbnail_path_ss'
 
     config.add_results_collection_tool(:sort_widget)
+    config.add_results_collection_tool(:algorithm_select)
```
tpendragon, line 122: Can we add this to the module somehow?

app/controllers/concerns/multiple_algorithms.rb
```diff
@@ -0,0 +1,25 @@
...
+    return unless search_algorithm_param && allowed_search_algorithms.include?(search_algorithm_param)
+
+    "#{search_algorithm_param}_search_builder".camelize.constantize
+  end
+
+  # When adding a new Ranking Algorithm the name will need to be added to this list before it can be utilized
+  # For example if you added a cats ranking algorithm, with a CatsSearchBuilder you would add "cats" to this list ["engineering","cats"]
+  # This is to make sure the user can not just execute any SearchBuilder in the system
+  def allowed_search_algorithms
+    ["online_content"]
```
tpendragon, line 19: Maybe this should be overridden in the catalog controller? I'm trying to think about how to make maintenance of this be as copy/paste as possible.

app/helpers/pulfalight_helper.rb
```diff
@@ -132,6 +132,15 @@ def regular_compact_breadcrumbs(document)
     )
   end
 
+  # Provides the search algorithm value for the UI component
+  def search_algorithm_value
```
tpendragon, line 136: Can we add default to allowed_search_algorithms and check that somehow?

tpendragon (changes requested):
Just a couple of things I think we should think about as we imagine this feature existing in multiple apps

app/controllers/catalog_controller.rb
eliotjordan: This is possible:
```

  def blacklight_config
    config = super
    config.add_results_collection_tool(:algorithm_select)
    config
  end
```

but it's not going to work out well. The location of the dropdown will vary depending on the app as each app will have different widgets enabled.

app/helpers/pulfalight_helper.rb
eliotjordan: Done.

app/controllers/concerns/multiple_algorithms.rb
eliotjordan: Done.

app/helpers/pulfalight_helper.rb
eliotjordan: Done

PR #1385, opened 2024-03-01
updated the schedule for removing temp db entries

### LINKED ISSUE #1250: Move search table expiration to whenever
To those of you with Blacklight apps that run on more than one server and use the blacklight_app role in Princeton Ansible - you may want to consider moving the cron job that the blacklight_app role creates into something generated by Whenever and run it on a single server. This is because Jane and Carolyn found that the cron job running simultaneously on multiple servers is creating database lock errors in postgres, which can cause problems for your application.

Here's a pair of PRs which hopefully resolve this for Orangelight
* [Adds the cron job to our "Whenever" configuration on Orangelight, so it runs on only one server](https://github.com/pulibrary/orangelight/pull/3627)
* [Removes the ](https://github.com/pulibrary/princeton_ansible/pull/4040)blacklight_app[ role from Orangelight, so that it does not re-create the cron job on all the servers](https://github.com/pulibrary/princeton_ansible/pull/4040)

Thanks to @maxkadel for the report

inieinie:
along with https://github.com/pulibrary/princeton_ansible/pull/6113 this closes #1250
ansible-generated cron jobs have been deleted from the servers

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -18,3 +18,11 @@ end
 every :day, at: "10:00 PM", roles: [:production_db] do
   rake "pulfalight:indexing:full DEFAULT_LOGGER=true"
 end
+
+every 1.day, at: "10:00pm", roles: [:db] do
+  logging_rake "blacklight:delete_old_searches", output: { error: "/tmp/guest_searches.log", standard: "/tmp/guest_searches.log" }
+end
+
+every 1.day, at: "10:30pm", roles: [:db] do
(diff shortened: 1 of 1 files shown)


config/deploy.rb
```diff
@@ -55,6 +55,19 @@
 set :pulfa_dir_path, fetch(:pulfa_dir_path, "/var/opt/pulfa")
 set :pulfa_collections, fetch(:pulfa_collections, %w[cotsen ea eng ga lae mss mudd rarebooks])
 
+# Default value for keep_releases is 5
+# set :keep_releases, 5
+
+desc "Write the current version to public/version.txt"
```
hackartisan, line 61: This should be removed; it duplicates code above

hackartisan (dismissed):

Issue #1292, opened 2023-07-18
Page should error with 404 not 500 when xml requested for non-existent record


hackartisan:
[PULFALight/production] ActionView::MissingTemplate: Missing template catalog/record_not_found, application/record_not_found with {:locale=>[:en], :formats=>[:xml], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :coffee, :jbuilder]}. Searched in:  * "/opt/pulfalight/releases/20230717150309/app/views"  * "/opt/pulfalight/shared/bundle/ruby/3.1.0/gems/devise-4.8.1/app/views"  * "/opt/pulfalight/shared/bundle/ruby/3.1.0/gems/blacklight_range_limit-7.9.1/app/views"  * "/opt/pulfalight/sha...

## Backtrace

    line 495 of [PROJECT_ROOT]/app/controllers/catalog_controller.rb: block in <class:CatalogController>

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/65922/faults/94082878)

## Note

This is returning a 500 for non-existent records period, not just for XML.

## To work on it

Write a failing test
Make it pass

## Sudden Priority Justification

Staff need to know if a record is missing rather than "broken" so they can diagnose what's going on and how to report it.

tpendragon:
This is sending a 500 even for HTML pages.

hackartisan:
The html page actually is returning a 404, I think the error page is maybe getting overwritten at the load balancer so I will look into that.

PR #1397, opened 2024-03-26
Add template for xml 404

hackartisan:
These were returning 500 because there was no template.

progresses #1292 

I also added some style classes into the 404 page to make it look a bit nicer:
Screenshot 2024-03-26 at 3 04 33 PM


tpendragon (approved):

hackartisan:
@tpendragon I added some styling if you'd like to re-review

PR #1400, opened 2024-03-28
updated the gem for sidekiq pro

### LINKED ISSUE #1265: Install Sidekiq Pro
Restarting the box loses jobs.

Remove this gem:
https://github.com/pulibrary/pulfalight/blob/main/Gemfile#L99

And add this to Gemfile:
https://github.com/pulibrary/figgy/blob/main/Gemfile#L103-L105

inieinie:
closes #1265


Gemfile
```diff
@@ -96,7 +96,9 @@ gem "popper_js"
 gem "rsolr", ">= 1.0", "< 3"
 gem "rubytree"
 gem "rubyzip", ">= 1.3.0"
-gem "sidekiq", "~> 6.5"
+source "https://gems.contribsys.com/" do
+  gem "sidekiq-pro"
+end
```
eliotjordan, line 101: Issue: While it will probably work this way, it would be better if you could move that block (lines 99-101) to [somewhere around here](https://github.com/pulibrary/pulfalight/blob/main/Gemfile#L81) with a space above and below.

eliotjordan (changes requested):
You'll need to run `bundle install` in the project root to update the `Gemfile.lock` file. You'll have to authorize sidekiq-pro before running since it's a gem we pay for. I'll put the instructions in Slack.

eliotjordan (approved):

PR #1411, opened 2024-05-01
Fix note propagation to Aeon

### LINKED ISSUE #1410: User-entered notes not passing from the cart to Aeon
Maybe we're no longer sending "Notes" as a field, or something's changed in Aeon.

## Acceptance Criteria

- [ ] When a note is entered in the cart it goes into Aeon's notes. You can tell if this will happen if the network request to AEON includes a form property "Notes" with the given property, as seen in Chrome's network tab.

## Replication Steps

- Click "Request This Box" on https://findingaids.princeton.edu/catalog/TC081_c006
- Type "Test Note" notes box in the cart
- Right click the box and "Inspect"
- Go to the Network tab in the top of the inspector that appears
- Click "Request 1 item"
- Scroll up to the top of the inspector window, click "logon", then in the right pick Payload
- Scroll down in the payload section - at the bottom should be Notes. There should be data there, there isn't:
Screen Shot 2024-05-01 at 8 06 41 AM

Without these notes our staff are unable to accurately support researchers.

tpendragon:
Closes #1410

### DIFF
--- a/app/javascript/components/RequestCart.vue
+++ b/app/javascript/components/RequestCart.vue
@@ -130,7 +130,7 @@
               width="expand"
               type="textarea"
               rows=2
-              v-model="note"
+              v-model:value="note"
             />
           </div>
           <lux-input-button type="submit" variation="solid" :disabled="requests.length == 0" block>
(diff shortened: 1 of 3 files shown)


hackartisan (approved):
The tests didn't catch this bug because they use `testing-library` which doesn't allow access to the vue component, so we can't emit the update event in the test. a future improvement would be to refactor the tests to use `vue-test-utils`

PR #1422, opened 2024-07-08
Add a context for errors in maintenance window.

tpendragon:
Work towards pulibrary/figgy#6394

This enables us to do a search in Honeybadger to exclude all maintenance items: `-context.maintenance_window:"true"`


sdellis (approved):
LGTM 👍

PR #1454, opened 2024-11-11
Only login to ASpace if we can't pull from cache.

### LINKED ISSUE #1453: Returning cached XML shouldn't login to ASpace.
## Background

We're getting pings from Lyrasis that ASpace is being hit by pulfalight quite often - every second or so. We should mitigate that if we can.

## Research done so far

It looks like the culprit is probably a bot scraping our XML. Most of the time our XML is in the cache though, so we shouldn't have to auth the ASpace client.

## Acceptance Criteria

- [ ] Returning cached XML shouldn't log in to ASpace

Lyrasis' servers can't handle the load we're putting on them by logging in so often.

tpendragon:
Closes #1453

### DIFF
--- a/app/services/aspace/client.rb
+++ b/app/services/aspace/client.rb
@@ -15,10 +15,15 @@ module Aspace
 
     def initialize
       super(self.class.config)
-      login
+    end
+
+    def login
+      # Only login if there's no token.
+      super unless @token
     end
 
     def ead_urls(modified_since: nil)
+      login
       output = repositories.each_with_object({}) do |repository, acc|
         config.base_repo = repository["uri"][1..-1]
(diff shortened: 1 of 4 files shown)


hackartisan (commented):
This seems good but I'm confused because the ticket talks about cached xml, and this PR memoizes the login token.

tpendragon:
> This seems good but I'm confused because the ticket talks about cached xml, but this PR memoizes the login token.

It was already caching XML. The ASpace Client used to login on instantiation, this delays it until it's needed - which for the methods which pull XML from the database isn't until after that cache misses.

hackartisan (approved):

Issue #1368, opened 2024-01-11
Make Collection Level Access restrictions badges clickable; and open the Access and Use Tab


ccleeton:
Users consistently try to click on badges thinking they are buttons.
For example when it shows Access Restrictions,, jump to the access page or the conditions governing access

## Acceptance Criteria

- [x] When going to https://findingaids.princeton.edu/catalog/MC247 or any collection page, clicking the "Restrictions May Apply" button will redirect them to https://findingaids.princeton.edu/catalog/MC247#access (or the appropriate access page for that collection)
- [x] The restrictions badge within search results is not a link. 
- [x] The restrictions badge within the Grouped by Collection is not a link.
- [x] The restrictions badge at the component level is not a link.
- [x] The restrictions badge at the series level is not a link. 

## First Step

Write a test.

ccleeton:
Example: https://findingaids.princeton.edu/catalog/MC247
At the top of the finding aid under the title and the online access green bar, make the "Restrictions may apply. See Access Note" clickable and open the Access & Use tab.

ccleeton:
Users consistently try to click the badge at the top of the finding aid. Users also consistently do not understand if they can actually view the material in person.

ccleeton:
I saw the ticket was closed but I am not seeing it active in the finding aid. Am I missing something?

hackartisan:
@ccleeton I've just deployed; it should be in place now!

faithc:
Hi @hackartisan and @Seanwarren77 -

I noticed that badges at lower levels of description are linking to the collection-level Conditions Governing Access note in the Access + Use tab. If it's possible to implement, badges associated with components should link to the Conditions Governing Access note applied at that component level.

For example, the badge at the [series level here](https://findingaids.princeton.edu/catalog/C1491_c1?onlineToggle=false) shouldn't link to the collection-level note, but rather to the note associated with that component. 

The [badge here](https://findingaids.princeton.edu/catalog/C1491_c00772-80460?onlineToggle=false) should link to the Conditions Governing Access note associated with the component, i.e., "Restricted for 75 years (until 2094)."

Please let Christa or I know if you have any questions.

Seanwarren77:
[...]

PR #1456, opened 2024-11-15
Added link to restrictions badge

Seanwarren77:
Screenshot 2024-11-19 at 9 42 11 AM

Now the link color is the same @tpendragon I also updated the test.
 
Ref #1368

### DIFF
--- a/spec/features/catalog_view_spec.rb
+++ b/spec/features/catalog_view_spec.rb
@@ -136,6 +136,7 @@ describe "viewing catalog records", type: :feature, js: true do
     it "lets someone click on the access restrictions badge", js: true do
       visit "/catalog/AC136_c2889"
       click_on "Restrictions may apply. See Access Note."
+      expect(page).to have_content("Access & Use")
     end
   end
   context "with a no-digital-content collection show page" do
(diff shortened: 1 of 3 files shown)


tpendragon:
> The color changed inside of the badge, do I need to change the color back to original color?

Yeah, we should have the same color.

spec/features/catalog_view_spec.rb
```diff
@@ -133,6 +133,10 @@
       expect(page).to have_selector(".document-access.review", text: "Restrictions may apply. See Access Note.")
       expect(page).to have_selector("#component-summary .document-access.restricted", text: "Restricted Content")
     end
+    it "lets someone click on the access restrictions badge", js: true do
+      visit "/catalog/AC136_c2889"
+      click_on "Restrictions may apply. See Access Note."
```
tpendragon, line 138: Can you add an assertion here so we know what it should be displaying after the click?

tpendragon (changes requested):

spec/features/catalog_view_spec.rb
```diff
@@ -133,6 +133,11 @@
...
+      expect(page).to have_content("Access & Use")
```
tpendragon, line 139: I think this is always there, https://findingaids.princeton.edu/catalog/MC247. Is there content we can check that is only there after you click the link, and not before?

tpendragon (changes requested):
This is looking good, just a small change.

tpendragon:
@Seanwarren77 Looking good. There's some code style problems (lint is failing) - you can fix them by running `bundle exec rubocop -a` and committing the changes that makes.

Issue #1447, opened 2024-10-08
Access restriction badge should appear on search results page


faithc:
### User story

As a user, it is important for me to see that materials are restricted from the search results page. The visibility of access restrictions should be robust. 

### Acceptance criteria

- [x] When viewing a restricted item in search results there should be a badge next to restricted items.

### Concrete example

A researcher does not know that any of these materials are restricted unless they click on the component. See for example, the search results listed [here](https://findingaids.princeton.edu/?group=true&f%5Bcollection_sim%5D%5B%5D=Toni+Morrison+Papers%2C+1908-2017+%28mostly+1970-2015%29&search_field=all_fields&q=Box+183). All of these materials are restricted. 

An example that isn't a component, but should have a badge: https://findingaids.princeton.edu/catalog/C1491_c84, search result: https://findingaids.princeton.edu/?f%5Bcollection_sim%5D%5B%5D=Toni+Morrison+Papers%2C+1908-2017+%28mostly+1970-2015%29&group=true&search_field=all_fields

### Implementation notes, if any

Adding "Restricted" badges to the search results page would likely be akin to current functionality that adds the "Has Online Content" badge to the search results page.

This should use the same logic as we use in the show page for the components.

These probably shouldn't be a link, because it would send them to the collection page and not the component page.

## First Step

Write a test. The Toni Morrison items in the test fixtures should include restricted items that can be used here.

PR #1460, opened 2024-11-27
Access restriction badge appears on search results page

Seanwarren77:
advances #1447


hackartisan (commented):
Nice!

There's one spot that's still a link but shouldn't be, it's in the grouped results for the entry below the top-level collection entry.

Screenshot 2024-11-27 at 12 26 09 PM

hackartisan:
The fact that there's one spot that's still a link makes me think there is probably a missing test, also

hackartisan (changes requested):
(changing to request changes; accidentally submitted my review as a comment)

faithc:
I think we could try that approach. Thanks.

ccleeton:
I agree with this approach!

hackartisan (approved):

PR #1467, opened 2024-12-05
restriction badge not a link at the series level

Seanwarren77:
advances #1368


sdellis (approved):
Looks great, @Seanwarren77 ! 👍

Issue #1488, opened 2025-01-13
Online only content toggle should have only one field label


Seanwarren77:
### Steps to recreate
https://wave.webaim.org/report#/https://findingaids.princeton.edu/catalog/WC064_c1

### URL
https://findingaids.princeton.edu/catalog/WC064_c1

### Issues
WAVE accessibility report rose errors due to multiple field labels. 

### Severity
High Severity

### WCAG Success Criterion
https://www.w3.org/TR/WCAG22/#labels-or-instructions

### WCAG Conformance level (A, AA, AAA)
AA

### User impact
Users of assistive technologies may not be given the appropriate label. 

### Recommended fix
Remove one of the labels for `tocOnlineToggle`

### Acceptance criteria
- [x] I can successfully use this feature with Voice Over in Safari.
- [x] I have checked this feature using one of the following browser extensions and confirmed that it does not add any level A or level AA issues:
    * axe dev tools
    * Lighthouse
    * Wave
- [ ] ...

PR #1493, opened 2025-01-23
Refactored online content toggle for accessibility 

Seanwarren77:
advances #1488
Screenshot 2025-01-30 at 9 34 31 AM

### DIFF
--- a/app/views/catalog/_show_collection.html.erb
+++ b/app/views/catalog/_show_collection.html.erb
@@ -27,11 +27,13 @@
         </div>
         <div class="toc-online-toggle al-online-content-icon">
           <%= blacklight_icon(:online) %>
-          <label for="tocOnlineToggle" class="online-toggle-label">Only show materials containing online content</label>
-          <div class="custom-control custom-switch">
-            <input type="checkbox" class="custom-control-input" id="tocOnlineToggle">
-            <label class="custom-control-label" for="tocOnlineToggle"></label>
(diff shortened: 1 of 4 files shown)


app/views/catalog/_show_collection.html.erb
```diff
@@ -27,11 +27,13 @@
...
-          <div class="custom-control custom-switch">
-            <input type="checkbox" class="custom-control-input" id="tocOnlineToggle">
-            <label class="custom-control-label" for="tocOnlineToggle"></label>
+          <label for="tocOnlineToggle" class="online-toggle-label">
+            <span> Only show materials containing online content </span>
+           <div class="toggle">
+              <input type="checkbox" id="tocOnlineToggle">
+              <span class="slider"></span>
           </div>
+          </label>
```
hackartisan, line 36: I expected the label tag to close before the toggle div starts, am I mistaken?

hackartisan (changes requested):
Just had a question about the markup. Also, can you add a screenshot?

Seanwarren77:
@hackartisan  I tried it the way you suggested and it did not work. So I added a test to make sure it stays working.

hackartisan:
Oh okay it looks like you can put the element inside the label. I've just never seen it that way. Here's more info: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

thanks for the extra test, good idea!

hackartisan (approved):

Issue #1490, opened 2025-01-15
Get up to speed with security backlog 


Seanwarren77:
There are currently 21 security warnings for pulfalight. Go through and apply them all. 

Acceptance criteria: 
- [ ] After this work is done there are zero security warnings. 

First step:
- [x] Go to the Security tab in Github
- [x] Start with the highest vulnerability

Seanwarren77:
Closed 20 out of 21, there is one security patch that we have decided not to address right now.

Issue #1076, opened 2022-05-13
Your Account Link


ccleeton:
Change the link of "Your Account" to https://lib-aeon.princeton.edu/aeon/aeon.dll
This will allow all users to login regardless of affiliation

sdellis:
@ccleeton I'm just confirming that you're talking about the "Your Account" login in the main navigation bar, correct? It appears to be linked to that address in all other instances I can see, but I agree that linking to the Aeon login makes much more sense.

ccleeton:
Yes the one in the nav bar because MANY of our outside researchers go right to that and try to log in and the current link only works for U members.
It also is much more intuitive to have a login at the top of the page instead of having to scroll "below the fold" to find the access.

eliotjordan:
What do we do about the Sidekiq dashboard? As far as I can tell, it's only Pulfalight-specific functionality that requires authentication using CAS. To be fair, it's not very secure as anyone with a NETID can view and interact with it.

hackartisan:
Good point. I'd vote to keep all the cas functionality in place for that. We don't need the login link, though, right? We can navigate directly to `/sidekiq`.

PR #1494, opened 2025-01-23
Update Omniauth-cas to 3.0.0

Seanwarren77:
advances #1490

Update: 
that omniauth stuff is only used to protect the /sidekiq page.
so the way to test whether the upgrade worked is to open an incognito window and go to findingaids-staging.princeton.edu/sidekiq (edited) 
more background here https://github.com/pulibrary/pulfalight/issues/1076
it looks like it's not working


Seanwarren77:
Blocked closing PR for now.

PR #1505, opened 2025-02-20
Upgrade vite for security

Seanwarren77:
Closes dependabot issue 194


bess (approved):

bess (changes requested):
I don't think this actually upgraded vite. Let's try again.

Issue #1515, opened 2025-02-24
Language should index multiple languages in an array


tpendragon:
### Summary or User Story

https://findingaids.princeton.edu/catalog/C0022_c0145/raw shows "language_ssm" as "English French". It should instead be `["English", "French"]`. This is currently causing https://github.com/pulibrary/figgy/issues/6468.

### Acceptance Criteria

- [x] Languages index as an array of every language.
- [x] The language display still looks reasonable on component pages.

### First step

Write a test.

PR #1517, opened 2025-02-28
Index multiple languages properly 

Seanwarren77:
Advances #1515

### DIFF
--- a/lib/pulfalight/traject/ead2_config.rb
+++ b/lib/pulfalight/traject/ead2_config.rb
@@ -373,6 +373,9 @@ to_field "language_sim" do |record, accumulator, _context|
     segments = value.split
 
     filtered = segments.reject { |e| e =~ /^[[:punct:]]/ }
+    # some languages, like "Greek, Modern" need to be concatenated
+    # if there is a comma, assume that the values should be joined together
+    # if there is not a comma, assume these are seperate values and should be treated as such
     values << if filtered.find { |e| /,/ =~ e }
                 filtered.join(" ")
               else
--- a/lib/pulfalight/traject/ead2_config.rb
+++ b/lib/pulfalight/traject/ead2_config.rb
@@ -376,17 +376,18 @@ to_field "language_sim" do |record, accumulator, _context|
     # some languages, like "Greek, Modern" need to be concatenated
     # if there is a comma, assume that the values should be joined together
     # if there is not a comma, assume these are seperate values and should be treated as such
-    values << if filtered.find { |e| /,/ =~ e }
-                filtered.join(" ")
-              else
-                filtered
-              end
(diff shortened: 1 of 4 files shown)


hackartisan (approved):

lib/pulfalight/traject/ead2_config.rb
```diff
@@ -371,15 +373,21 @@
     segments = value.split
 
     filtered = segments.reject { |e| e =~ /^[[:punct:]]/ }
```
tpendragon, line 375: This is very strange to me. I don't understand why we need to split on punctuation and figure out what's a language when the XML is already split out in the EAD. I wonder if we need to pull it from `/ead/archdesc/did/langmaterial/language` instead, so it doesn't combine all the languages?

Alternatively I think there's a way get the nodes out raw so we can look at just each individual language tag.

Can you spend a little time looking into that? I'm scared parsing this text out is gonna lead to strange bugs.

tpendragon (changes requested):

bess (changes requested):
Let's look at langmaterial/language instead and see if we can make this language parsing less fragile

bess:
@tpendragon Yeah, I agree. Thank you for advocating for the more sustainable solution instead of the quick fix. I talked to @regineheberlein and she agrees we should refactor the language indexing code to look at langmaterial/language instead. 

She also looked at that weird case I showed you where there is no language string, and maybe it is a bug in ASpace? She will follow up with more testing.

regineheberlein:
You all found a blip with our migration to ArchivesSpace! It's somewhat entertaining in its capriciousness and you can read all about what I found out in the ticket linked above.

I'll fix the data issue. The language(s) of materials will by default export like so:

```
    <langmaterial>
      <language langcode="eng">English</language>
,       <language langcode="fre">French</language>
,       <language langcode="rus">Russian</language>
,       <language langcode="ger">German</language>
.    </langmaterial>
```

As Trey says, for pulfalight you could just grab each language element and concatenate the string values together with a delimiter of your choice--perhaps a semicolon since some of them, like "Greek, Modern" already have commas. Or you could even just grab the 3-letter codes and process those instead.

(Note that the serialization above already HAS a comma and a period in between and after the `language` elements, which to me is super weird and I would ignore, but hey, it's there.)

regineheberlein:
@Seanwarren77 @bess 
Those should be fixed, C1408 langmaterial EAD now gives me
```
<langmaterial>
      <language langcode="gre">Greek, Modern (1453-)</language>
,       <language langcode="eng">English</language>
,       <language langcode="fre">French</language>
,       <language langcode="bul">Bulgarian</language>
.    </langmaterial>
```
--you should be ok to regenerate your fixture.

PR #1529, opened 2025-03-19
Updated has online content to has online material

Seanwarren77:
Advances #1524 
Screenshot 2025-03-19 at 10 26 12 AM

PR #1531, opened 2025-03-19
Make online content badge clickable

### LINKED ISSUE #1521: Some Online Material button on record pages should lead to search results
### Summary (Product Owner)

The "some online content" badge on a show page should lead to search results for online content within that series.

Example: https://findingaids.princeton.edu/catalog/C1491_c1?onlineToggle=false

Image

Extracted from #1374 

### Implementation Notes

I suggest adding `config.add_facet_field "parent_unittitles_ssm", label: "Series", show: false` to catalog controller so you can facet by it, then you can make the badge go to a link like `https://findingaids.princeton.edu/catalog?f[parent_unittitles_ssm][]=Writings,%201917-2017%20April%2011&f[has_direct_online_content_ssim][]=online`

### Acceptance Criteria

- [ ] Clicking `Some Online Content` in https://findingaids.princeton.edu/catalog/C1491_c1?onlineToggle=false goes to `https://findingaids.princeton.edu/catalog?f[parent_unittitles_ssm][]=Writings,%201917-2017%20April%2011&f[has_direct_online_content_ssim][]=online`, which will show only online content within that series.
- [ ] Pages with "Online Content" should not be a link.

### First Step

Update https://github.com/pulibrary/pulfalight/pull/1531 to finish this ticket.

bess:
When a user clicks the online content badge, the system performs a search for the online content within that component.

Closes #1521

### DIFF
--- a/spec/features/catalog_view_spec.rb
+++ b/spec/features/catalog_view_spec.rb
@@ -103,8 +103,8 @@ describe "viewing catalog records", type: :feature, js: true do
       it "renders a place for javascript to enter a viewer" do
         expect(page).to have_css("#readingroom")
       end
-      it "displays 'Has Online content' as a link at the component level" do
-        expect(page).to have_selector("div.document-attributes > a > .document-access.online-content", text: "HAS ONLINE MATERIAL")
+      it "displays 'Has Online Material' at the component level" do
+        expect(page).to have_selector("div.document-attributes > .document-access.online-content", text: "HAS ONLINE MATERIAL")
       end
       it "displays 'Has Online Material' at the collection level" do
         visit "/catalog/MC221"
@@ -116,6 +116,15 @@ describe "viewing catalog records", type: :feature, js: true do
       it "displays 'Has Online Material' in its contents list" do
         visit "/catalog/MC221_c0092"
         expect(page).to have_selector("#component-summary .child-component-table td div.online-direct-content", text: "HAS ONLINE MATERIAL")
+      end
+    end
+
+    context "with a component with children, some of which have online material" do
+      it "lets the user search for online materials" do
+        visit "/catalog/C1491_c3"
+        click_on("SOME ONLINE MATERIAL")
+        expect(page).to have_content "1 entry found"
+        expect(page).to have_content "Box 1, Folder 1"
       end
     end
   end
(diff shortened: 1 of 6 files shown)


Seanwarren77 (approved):

hackartisan (changes requested):
Merging #1529 made this PR have conflicts, unfortunately, so it will need to be rebased.

But I am sort of confused about this. I think there are a few scenarios and I'm not sure my understanding of what should happen is correct, here is what I think we want:

- series-level component record (i.e. non-terminal node) for which some or all descendants have online material: online material badge is a link
- component record (i.e. terminal node) which itself has online material: I don't think it the badge needs to be a link because the viewer will show right below it at the top of the metadata area. example record: https://findingaids.princeton.edu/catalog/C1491_c4?onlineToggle=false -- if the badge is a link it will take you to a search form showing all the children of its parent that have online material. This seems to lead a user further away from the online content they were presumably looking for, and also potentially lead them into a neverending loop (if they select the same record again, and then click that link again, etc.)

hackartisan:
If my comment above reflects what this PR does, then I maybe misread the tests? And more comments on the tests would be helpful

tpendragon:
> If my comment above reflects what this PR does, then I maybe misread the tests? And more comments on the tests would be helpful

I'm hearing "only records that display "Some Online Material" should have a link", and since there's a branch like that it might be easier to handle that code somewhere like https://github.com/pulibrary/pulfalight/blob/109146c49233d812e936d3280d8aba324898e111/app/presenters/online_content_badge.rb where you have access to all that.

spec/features/catalog_view_spec.rb
```diff
@@ -315,7 +315,12 @@
           expect(page).to have_selector "dd.blacklight-collection_bioghist_ssm", text: /Noël Riley Fitch was born on December 24, 1937/
         end
       end
-
+      it "lets the user search for online materials" do
```
hackartisan, line 318: This is in the collection page context, but it visits a component page. Could we move it? Maybe near the one I added in #1566?

hackartisan (changes requested):
This looks good, just one suggestion about the test

spec/features/catalog_view_spec.rb
eliotjordan: 👍 Done.

hackartisan (approved):

Issue #1528, opened 2025-03-19
Error on startup


bess:
Application currently gives an error on startup.  

<img width="1712" alt="Image" src="https://github.com/user-attachments/assets/35c9192c-54a6-4f38-95e3-3ae7db25021f" />

@Seanwarren77 We had fixed this in a PR that didn't get merged. Let's fix it in a separate PR this time.

PR #1533, opened 2025-03-20
Changed config.cache_classes to equal false

Seanwarren77:
Advances #1528


tpendragon (changes requested):
Can you provide some reasoning as to why?

My understanding is that `cache_classes` improves test run time, because it doesn't have to dynamically load and look up classes every time it runs into one. This is usually good for dev because we change the code while the dev server's running, but we don't usually change the code in the middle of a test run.

tpendragon (approved):
Nope, nevermind, I read the issue.

Spring, gotta love it.

PR #1541, opened 2025-03-26
Proxy ASpace API interactions in rake tasks through a worker box.

tpendragon:
This means we don't have to bother Lyrasis with requests to change our
IP ranges anymore through VPN.

To test you'll have to pull this down, run `yarn install`, and then you should be able to connect to VPN, delete one of the fixtures in `spec/fixtures/aspace/generated/mss/*`, and run `bundle exec rake pulfalight:fixtures:refresh_aspace_fixtures`


sdellis (approved):
LGTM! 👍

Issue #1516, opened 2025-02-24
Date Range Facet is Broken


sdellis:
### Expected behavior
Users should be able to set limits on the Date Range facet by dragging either the begin or end date.

<img width="296" alt="Image" src="https://github.com/user-attachments/assets/57c5a40b-e3fd-4718-b54e-5169d3b95362" />

### Actual behavior
There is an opaque grey div with a class of `slider slider-horizontal` that obscures the chart and prevents user interaction. Perhaps it needs to be made transparent? I'm not sure what use it serves but may be part of the widget.

<img width="301" alt="Image" src="https://github.com/user-attachments/assets/9852ced2-fb2b-44cd-8fe6-f20d2669c81f" />

### Steps to replicate
Search for something and then try to use the Date Range facet.

### Impact of this bug
Users can enter dates manually, but it is not intuitive and it doesn't look good.

### Acceptance criteria
- [ ] If this change touches any of the files referenced in `.rubocop_todo.yml` create a new ticket to address the rubocop  todo.

PR #1547, opened 2025-03-28
Remove date range color configuration.

tpendragon:
Finishes fixing #1516. Thanks @seanaery

Screen Shot 2025-03-28 at 8 22 47 AM


sdellis (approved):
Always nice when deleting code fixes the issue! LGTM 👍

PR #1556, opened 2025-04-29
Upgrade vite to 6.x

### LINKED ISSUE #1555: Upgrade to vite 6.x
It's blocking an esbuild update that dependabot wants. And we're going to need to do it eventually either way.

hackartisan:
closes #1555


sdellis (approved):
Looks good!

PR #1565, opened 2025-05-02
Initial implementation

### LINKED ISSUE #769: Users should be able to right click and open a new tab from the Table of Contents
## Summary

When users try to open a new tab from the ToC, the existing page loads in the new tab, not the page they wanted. The use case for this behavior is to be able to get information on a variety of components. The ToC refreshes after every click, sometimes making it difficult to preserve context in series with a long list of similarly named component. Even if this was fixed, the new tab behavior should work and will help staff and researchers evaluate materials. 

Example: https://findingaids.princeton.edu/catalog/C0159_c000001?onlineToggle=false Right click and open in new tab the second ToC link.

## Acceptance Criteria

- [ ] When right clicking a link from the table of contents and opening it in a new tab it should open that component, not the page it's currently on.

## Implementation Details

Our table of contents is setting every hyperlink to have an `href` of `#` - those should be real links, even if our javascript is capturing them, so that middle clicks work.

## First Step

Write a test to ensure that the ToC links are real links, not all pointing to `#`.

eliotjordan:
Closes #769 

**This PR was created, in part, with the assistance of an LLM. Please review carefully.**

### DIFF
--- a/spec/features/toc_spec.rb
+++ b/spec/features/toc_spec.rb
@@ -7,7 +7,6 @@ describe "Table of Contents", type: :feature, js: true do
     it "has real href links" do
       visit "/catalog/MC221"
 
-      expect(page).to have_css("#toc")
       expect(page).to have_css("#toc .jstree-node")
 
       within("#toc") do
@@ -22,6 +21,22 @@ describe "Table of Contents", type: :feature, js: true do
           expect(href).to include("/catalog/")
         end
       end
+    end
+  end
+
+  describe "online collection toggle switch" do
+    it "has a toggle switch for showing materials containing online material", js: true do
+      visit "/catalog/C1491"
+      expect(page).to have_content "Working Files, 1955-2018"
+      find(".toggle > span").click
+      expect(page).not_to have_content "Working Files, 1955-2018"
+    end
+  end
+
+  describe "components with a viewer" do
+    it "displays an icon in the table of contents", js: true do
+      visit "/catalog/MC221_c0094"
+      expect(page).to have_selector "li#MC221_c0094 .online-direct-content"
     end
   end
 end
--- a/spec/features/toc_spec.rb
+++ b/spec/features/toc_spec.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe "Table of Contents", type: :feature, js: true do
+  describe "links in the table of contents" do
+    it "has real href links" do
+      visit "/catalog/MC221"
+
+      expect(page).to have_css("#toc")
+      expect(page).to have_css("#toc .jstree-node")
+
+      within("#toc") do
+        links = page.all("a")
+
+        expect(links[0]["href"]).to include("/catalog/MC221_c0001")
+        # Check that none of the links have href="#"
+        links.each do |link|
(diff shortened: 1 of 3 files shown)


spec/features/toc_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+require "rails_helper"
+
+RSpec.describe "Table of Contents", type: :feature, js: true do
+  describe "links in the table of contents" do
+    it "has real href links" do
+      visit "/catalog/MC221"
+
+      expect(page).to have_css("#toc")
+      expect(page).to have_css("#toc .jstree-node", wait: 10)
```
tpendragon, line 11: Why the `wait`? Can't it just be `expect(page).to have_css("#toc .jstree-node")`?

spec/features/toc_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+        # Make sure we have at least one link
+        expect(links.length).to be > 0
+
+        # Check that none of the links have href="#"
+        links.each do |link|
+          href = link["href"]
+          expect(href).not_to eq("#")
+          expect(href).not_to be_nil
+          expect(href).to include("/catalog/")
```
tpendragon, line 24: Can we test that a specific link is there?

tpendragon (changes requested):

spec/features/toc_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Table of Contents", type: :feature, js: true do
```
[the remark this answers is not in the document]
eliotjordan: I did see the catalog view spec file when I was working up the Claude output. There is only one very basic (#toc exists) test there. I think having a separate feature test file for the TOC is a good idea.

spec/features/toc_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe "Table of Contents", type: :feature, js: true do
```
tpendragon, line 5: I understand you decided having a TOC feature spec by itself is a good idea. I can get behind that.

If so can we move the other toc specs here?

https://github.com/pulibrary/pulfalight/blob/main/spec/features/catalog_view_spec.rb#L118-L120

https://github.com/pulibrary/pulfalight/blob/main/spec/features/catalog_view_spec.rb#L91-L94

https://github.com/pulibrary/pulfalight/blob/main/spec/features/catalog_view_spec.rb#L319-L324

tpendragon (changes requested):

spec/features/toc_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  describe "links in the table of contents" do
+    it "has real href links" do
+      visit "/catalog/MC221"
+
+      expect(page).to have_css("#toc")
```
tpendragon, line 10: Do we need this line if we have the next line?

spec/features/toc_spec.rb
eliotjordan: Done.

tpendragon (approved):

Issue #1523, opened 2025-03-15
Contents list should have a badge for "online material" or "some online material"


tpendragon:
### Summary (Product Owner)

Folks were browsing the content list in finding aids trying to find things with online content. There should be a badge to show which ones have some.

<img width="984" alt="Image" src="https://github.com/user-attachments/assets/20bb07a4-8042-4c64-b82c-faecb1e5fae5" />

Those badges should behave the same as the badges on record pages.

Example: https://findingaids.princeton.edu/catalog/C1491_c1?onlineToggle=false

Extracted from #1374 

We're not sure if we should display the whole badge or a smaller icon like what's in the table of contents. Try what's in your heart, and we can iterate.

### Acceptance Criteria

In the content list: 
- [ ] Series which have components that have online content should have a "some online content" badge. Clicking it should go to a search result like #1521
- [ ] Components which have online content (has_direct_online_content_ssim:true) should have an online content badge. Clicking it should go that component.

### First Step

Write a test.

PR #1566, opened 2025-05-05
Add an "online' column to contents list

hackartisan:
For consideration, this is the simplest implementation to satisfy #1523. It results in a small footprint for the badge and maximal context for its meaning.

Screenshot 2025-05-05 at 11 41 44 AM


hackartisan:
@tpendragon I added a test, let me know what you think

tpendragon (approved):

PR #1568, opened 2025-05-06
Open TOC to first component with online content

### LINKED ISSUE #1522: Clicking "Only show materials with online content" toggle in table of contents should expand to the first component
### Summary (Product Owner)

Image

Users should see something happen when they toggle to only show online content in the table of contents, such as in https://findingaids.princeton.edu/catalog/C1491_c1?onlineToggle=false.

We'd like to try expanding that table of contents far enough to see one component with the fully green icon when you toggle it on.

Extracted from #1374 

### Implementation Notes

On the UI side, the logic for toggling online content lives here:
https://github.com/pulibrary/pulfalight/blob/main/app/javascript/pulfalight/toc.es6.

The toggle will most likely need to trigger an ajax call to the [TOC controller](https://github.com/pulibrary/pulfalight/blob/main/app/controllers/toc_controller.rb) which will call the [TableOfContentsBuilder](https://github.com/pulibrary/pulfalight/blob/main/app/services/table_of_contents_builder.rb) with a new parameter (online_content: true). This will generate a tree where the nodes are expanded to show the first component with online content. 

### Acceptance Criteria

- [ ] When the table of contents is collapsed, toggling to view only online content will hide all non-online content and expand the table to the first possible component with online content. If you toggle it back off, it displays the non-online content it hid before, but doesn't collapse or expand anything.

### First Step

Probably fiddle around and read the jstree documentation, maybe write a test.

eliotjordan:
Closes #1522 

https://github.com/user-attachments/assets/8d745fd0-dadd-412b-becd-0549312cdde9

### DIFF
--- a/app/javascript/pulfalight/toc.es6
+++ b/app/javascript/pulfalight/toc.es6
@@ -5,6 +5,11 @@ export default class TocBuilder {
     // have the correct selected node id. The toc element itself is permanent (not
     // updated by turbolinks), and is instead updated by event-triggered javascript.
     this.dataElement = $(`${element}-data`)
+
+    // Setup the initial online toggle value if it's missing
+    if (this.getOnlineToggleValue() === null) {
+      this.getOnlineToggleValue('false')
+    }
   }
 
   get expanded() {
@@ -22,7 +27,7 @@ export default class TocBuilder {
       url = `${url}&expanded=true`
     }
 
-    if (this.getOnlineToggleParam()) {
+    if (this.getOnlineToggleValue()) {
       url = `${url}&online_content=true`
     }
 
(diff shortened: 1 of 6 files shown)


app/javascript/pulfalight/toc.es6
```diff
@@ -22,6 +27,10 @@ export default class TocBuilder {
       url = `${url}&expanded=true`
     }
 
+    if (this.getOnlineToggleValue()) {
+      url = `${url}&online_content=true`
```
sdellis, line 31: Do we need the URL query string here? The cookie should take care of that, right? Unless we want to be able to set the toggle via URL, but I don't see where that value would carry through.

eliotjordan: This is the url for the AJAX request. Not the url the user sees in the browser. It is needed.

PR #1576, opened 2025-05-07
Preserves online ToC toggle state by collection

### LINKED ISSUE #1290: Persisting the online toggle should not use the query params
## Summary

When clicking the online toggle it adds a query parameter when you click an item in the table of contents, that looks like this: https://findingaids.princeton.edu/catalog/AC122_c1081?onlineToggle=true

Archivists often copy and paste that URL for patrons and frequently have to remove the parameter, as they don't want that filter to apply to the person they're giving it to. If there's not online content, and it's there, it may be confusing for them to see it.

## Acceptance Criteria

- [ ] Clicking the toggle filters the page to online resources only.
- [ ] Clicking an element after toggling leaves the table of contents toggled, but doesn't add anything to the URL.
- [ ] Opening another collection should not have the toggle set.
- [ ] Opening this collection in another tab should have the toggle set (not a blocker, we just think this is easier, if it's not, don't do that.)

## First Step

Add a test.

## Technical Tips

https://github.com/pulibrary/pulfalight/pull/1295 - This has a PR that uses cookies instead, will help the implementation here.

sdellis:
Closes #1290

### DIFF
--- a/app/javascript/pulfalight/toc.es6
+++ b/app/javascript/pulfalight/toc.es6
@@ -1,6 +1,8 @@
 export default class TocBuilder {
   constructor(element) {
     this.element = $(element)
+    this.online_toggle_state = new Map()
+    this.collection = null
     // We need a separate data element that is updated by turbolinks so we always
     // have the correct selected node id. The toc element itself is permanent (not
     // updated by turbolinks), and is instead updated by event-triggered javascript.
@@ -22,7 +24,7 @@ export default class TocBuilder {
       url = `${url}&expanded=true`
     }
 
-    if (this.getOnlineToggleValue()) {
+    if (this.getOnlineToggleValue(this.collection)) {
       url = `${url}&online_content=true`
     }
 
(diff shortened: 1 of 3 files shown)


app/javascript/pulfalight/toc.es6
```diff
@@ -59,6 +59,15 @@ export default class TocBuilder {
   }
 
   setupToggleElement() {
+    var prev_collection=localStorage.getItem("collection");
+    var current_collection=$('#document').data('document-id').split("_").at(0);
+
+    // reset toggle state if the user has moved to a new collection
+    if(prev_collection !== current_collection) {
```
tpendragon, line 66: If I navigate between collections between tabs, is this gonna result in strange behavior?

tpendragon (changes requested):

app/javascript/pulfalight/toc.es6
sdellis: Yes, you're correct @tpendragon. Thanks for catching that. I think I need to store the concatenated value of the the browser tab id and the collection id as elements of an array of current tabs in a LocalStorage variable.

tpendragon: Oh wow I didn't even know tab ids were a thing. That's cool.

app/services/table_of_contents_builder.rb
```diff
@@ -58,6 +58,7 @@ def expand?(component)
     if component["id"] == document["id"]
       true
     elsif @online_content && component["has_online_content_ssim"]&.first == "true" && !@found_online_content
+      # component["has_online_content_ssim"]&.first turns false before it hits the online component - shaune
```
tpendragon, line 61: I don't really understand what this means.

spec/features/toc_spec.rb
```diff
@@ -53,6 +53,34 @@
...
+        page.execute_script('window.open("/catalog/MC221", "_blank");')
+      end
+      within_window new_tab do
+        expect(page).to have_content "Series 1: U.S. diplomacy career, 1900-1978"
+      end
+
+      # Opening a new tab containing the previous collection should preserve that collections's online toggle state
+      page.refresh
+
+      # Displays online content
```
tpendragon, line 79: Maybe a comment that says you're doing this to make sure that the page rendered after refresh.

tpendragon (changes requested):
Just a couple small questions. Looks amazing!

app/services/table_of_contents_builder.rb
sdellis: That was a note for @eliotjordan  as he was reviewing since I noticed strange behavior that was probably related to the bug in the JS code.  Meant to delete it... it's been removed.

tpendragon (approved):

PR #1580, opened 2025-05-09
Show all container types in Summary Storage Note

### LINKED ISSUE #1491: Storage notes don't display for all container types
### Expected behavior
Storage notes should display for all container types present in finding aid data.

### Actual behavior
Container information that's associated with the Storage note isn't being populated or displaying at all in cases where containers are identified as items or volumes rather than boxes. See, for example, https://findingaids.princeton.edu/catalog/AC297_c742 and https://findingaids.princeton.edu/catalog/MC006_c00068 where collection materials are indicated as items or volumes as opposed to boxes. 

```<container altrender="scarcpph" encodinganalog="NBox" id="4e1f0d4e21d48ed64c2d48d529f8cbd5" label="Mixed Materials [32101054257280]" type="item">1978 Brown (Defense 1)</container>```

```<container altrender="scamudd" encodinganalog="Mudd OS height" id="dc334d21d9fc2e7219478422a1d2d059" label="Mixed Materials [32101105971582]" type="volume">97</container>```

For these collections, none of the information for these items/containers displays in the Storage note.  

Related to #1302 

### Impact of this bug
Staff and researchers don't know in which location containers are located. 

### Acceptance criteria
- [ ] Storage notes should display on Collection Overview for all container types (Box, Item, Volume, Case, Folder, and Mixed Materials). 
  - [ ] The storage note for [Bernard M Baruch Papers](https://findingaids.princeton.edu/catalog/MC006) should have a bullet for "Volumes" which includes the "located in" value from [MC006_c00068])https://findingaids.princeton.edu/catalog/MC006_c00068), "Volume 97, Volume 98, Volume 99, Volume 100, Volume 101"
  - [ ] The storage note for [Princeton Football Films Collection](https://findingaids.princeton.edu/catalog/AC297) should have a bullet for "Items" which includes the "located in" value from [AC297_c742](https://findingaids.princeton.edu/catalog/AC297_c742), "Item 1978 Brown (Defense 1)"

Output Example:

```
* Location (locationcode):
     * Boxes: 1-10;
     * Items: 1-2
     * Volumes: 1-4
```

Out of scope: consolidation of those notes into ranges, which is covered by #1302 

### First Step
Write a test. 

## Implementation Details

This probably means adjusting https://github.com/pulibrary/pulfalight/blob/main/app/presenters/summary_storage_note_presenter.rb and https://github.com/pulibrary/pulfalight/blob/main/lib/pulfalight/traject/ead2_component_config.rb#L359-L370 and https://github.com/pulibrary/pulfalight/blob/main/lib/pulfalight/traject/ead2_config.rb#L109-L121

We have to make sure that we're only displaying containers that are top containers. We've done this before in indexing, but I'm not sure where. You'll have to do some research to determine which containers are top containers. Info from our session says:

1. Things with a barcode (maybe not that)
2. Things in container_information_ssm?

Whatever code we're using now for requesting, where we determine which container goes on the call slip, is working.

Something sent in Slack:

For instance https://findingaids.princeton.edu/catalog/C1573_c9. It has two containers, Box P-000090 and Folder 8, folder 8 is a child of the box so it's not a "top" container. If you look in https://findingaids.princeton.edu/catalog/C1573_c9/raw in container_information_ssim there are two entries - the folder has a parent key that points to the ID of the box. Top Containers are all containers that don't have a parent key.

regineheberlein:
This PR changes the way the summary storage note is indexed, and ensures that display can handle both the old index format and the new. clean up work will be needed after this is deployed and indexed, #1581 has been created for that

closes #1491

### Screenshots
fixture resource indexed with a folder container type:
Screenshot 2025-05-09 at 3 02 30 PM

new appearance uses a description list rather than a nested unordered list, because we had more of a key / value data shape and also because axe-core did not allow ul tags within a ul.
Screenshot 2025-05-09 at 3 00 41 PM


tpendragon (approved):
Looks great, good work! The comments were well done too, quite easy to follow.

PR #1587, opened 2025-05-14
Sync cart items between tabs

### LINKED ISSUE #1554: Request Cart Issues when using multiple browser tabs
### Description of Bug
The Request Cart does not share its data across tabs so a user who is trying to add items to their cart via multiple browser tabs ends up seeing items that were deleted in other tabs, or missing items that were added in other tabs. Tabs are frequently used while doing research, so this is a bug that should be resolved. 

### How it Should Behave
A user opens two components (A and B) in two tabs (A and B). The user adds Component A to their request cart via Tab A. When the user switches to Tab B and adds Component B to their request cart, both Components should be in the Request Cart. 

### Acceptance criteria

When two tabs are open, both with the request cart open:

- [ ] Requesting an item in tab A, then looking at tab B, shows items in both tab A/B
- [ ] Requesting an item in tab B, then looking at tab A, shows items in both tab A/B
- [ ] Removing an item from either tab removes it from the other tab.
- [ ] If you close the cart in tab A, request an item in tab B, and then open the cart in tab A, the new item should also be in tab A.

### First Step

Do some research into how to do this.

### Technical Tips

ActionCable will probably be useful for this.

Try to avoid saving this state on the server if we can.

hackartisan:
Use the local storage event and vue `replaceState` to sync cart between tabs

closes #1554

### DIFF
--- a/app/javascript/pulfalight/pulfalight_loader.es6
+++ b/app/javascript/pulfalight/pulfalight_loader.es6
@@ -22,7 +22,7 @@ export default class {
     this.setup_lib_cal_hours()
     this.setup_form_modal()
     this.setup_media_queries()
-    this.setup_cart_sharing()
+    this.setup_cart_sync()
   }
 
   setup_lib_cal_hours() {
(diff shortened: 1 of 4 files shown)


sdellis (approved):
Wow, great solution and simpler than using the Broadcast API. 👍

Issue #1621, opened 2025-12-03
Send feedback emails via the libanswers api instead of using mailers


eliotjordan:
## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/65922/faults/124747779)

### Summary (Product Owner)

We should switch to use the libanswers API, which would let us remove sending emails (or at least we could move to the new mail server)

### Priority recommendation (Product Owner)

- [ ] asap (add sudden priority justification below)
- [x] within the next 3 weeks (add sudden priority justification below)
- [ ] PO will prioritize

### Sudden Priority Justification (Product Owner)
Emails are being lost, or would be if we weren't using honeybadger stampede to manually create them.
### Acceptance Criteria (Product Owner / Developers)

This work is complete when:

- [ ] Suggest a correction without a from email works in production.

### First Step (Developers)

Go find what they've done in Orangelight to use the libanswers API.

### Implementation Notes (Developers)

Relevant links:

springshare docs about libanswers api: https://ask.springshare.com/libanswers/faq/1669
orangelight PR: https://github.com/pulibrary/orangelight/pull/4073/files
orangelight ansible PR: https://github.com/pulibrary/princeton_ansible/pull/5206/files
cool spam protection thingie in orangelight: https://github.com/pulibrary/orangelight/blob/main/app/forms/concerns/honeypot.rb#L9

DPUL-C implementation: https://github.com/pulibrary/dpul-collections/issues/727

PR #1628, opened 2025-12-23
Use LibAnswers API instead of email

sdellis:
refs #1621

### DIFF
--- a/app/models/suggest_a_correction_form_submission.rb
+++ b/app/models/suggest_a_correction_form_submission.rb
@@ -47,9 +47,9 @@ class SuggestACorrectionFormSubmission
   end
 
   def token
-    OAuthToken.find_or_create_by({ 
-      service: "libanswers",
-      endpoint: "https://faq.library.princeton.edu/api/1.1/oauth/token" 
-    }).token
+    OAuthToken.find_or_create_by({
+                                   service: "libanswers",
+                                   endpoint: "https://faq.library.princeton.edu/api/1.1/oauth/token"
+                                 }).token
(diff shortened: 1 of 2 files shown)


spec/forms/suggest_a_correction_form_spec.rb
```diff
@@ -27,20 +28,57 @@
...
+        )
+      end
+    end
+
+    context "when the location code is engineering" do
+      it "sends an email" do
+        form = described_class.new(valid_attributes.merge({ "location_code" => "engineering library" }))
+
+        form.submit
+        # expect(ActionMailer::Base.deliveries.length).to eq 1
```
hackartisan: we should remove this line

app/models/suggest_a_correction_form_submission.rb
```diff
@@ -0,0 +1,54 @@
...
+  def send_to_libanswers
+    Net::HTTP.post uri, body, { Authorization: "Bearer #{token}" }
+  end
+
+    private
+
+  attr_reader :name, :email, :box_number, :location_code, :context, :user_agent
+
+  def body
+    @body ||= data.to_a.map { |entry| "#{entry[0]}=#{entry[1]}" }.join("&")
```
eliotjordan: issue: While not critical, all of the methods in this class are prematurely memoized. I.e. they do not need the `@body ||=` assignment. `uri` and `token` methods as well. They are only referenced once on line 18.

eliotjordan (changes requested):

PR #1631, opened 2026-01-09
Ensure asdf Ruby version matches CI and production

eliotjordan:
There was a concern that rubocop didn't run correctly on 3.2.8, but it was probably a local setup issue:

image


sdellis (approved):
LGTM 👍

Issue #1636, opened 2026-01-29
Ensure form submissions still submit to libanswers


hackartisan:
### Summary or User Story

DCAS discovered that their submissions to the libanswers API were no longer going through (see pulibrary/orangelight#5558). We should check ours and if they're not working, fix them.

### Acceptance Criteria

- [ ] If a submission is attempted and fails to go through the API, we get a legible error in honeybadger that contains the submission data we'd require to succeed on retry or manually create
- [ ] form submissions generally work, creating a libanswers ticket via the API

### First step

look at PRs that resolve orangelignt issue linked above: https://github.com/pulibrary/orangelight/pull/5577/changes

Issue #1630, opened 2026-01-09
Engineering Library staff can use LibAnswers to receive and respond to finding aid feedback forms


sdellis:
### Summary (Product Owner)
The Engineering Library is currently not using LibAnswers to manage correspondence around the Finding Aids collections. They plan to move to LibAnswers, and once they are ready, we should make sure that messages are routed there.  Once this ticket is complete and once #1629 is complete we can stop using the Mailer to deliver the messages. 

Note: LibAnswers has a notification mechanism so that staff can still receive notifications of new messages through email. 

### Priority recommendation (Product Owner)

- [ ] asap (add sudden priority justification below)
- [ ] within the next 3 weeks (add sudden priority justification below)
- [ ] PO will prioritize

### Acceptance Criteria (Product Owner / Developers)

This work is complete when:

- [ ] The Engineering Library staff can use LibAnswers to correspond with patrons who message through the Suggest a Correction and Ask a Question forms. 

### First Step (Developers)
Look at how this is done in #1628

faithc:
@sdellis - is this ticket still in progress?

Issue #1629, opened 2026-01-09
Use the LibAnswers API for the Ask a Question form


sdellis:
### Summary (Product Owner)
The Suggest a Correction form currently uses the LibAnswers API instead of email due to resiliency and a smaller infrastructure footprint. We want all of the forms to use this API so we can get rid of the mail server. 

Note: The Engineering Library is not currently using LibAnswers and relies on email, but that repository will be moving there eventually as well. 

This is a follow up to #1621. 

### Priority recommendation (Product Owner)

- [ ] asap (add sudden priority justification below)
- [ ] within the next 3 weeks (add sudden priority justification below)
- [ ] PO will prioritize

### Sudden Priority Justification

We need to get off the old mail server.

### Acceptance Criteria (Product Owner / Developers)

This work is complete when:

- [ ] The Ask a Question form uses the LibAnswers API for message delivery for Special Collections locations. 

### First Step (Developers)
Look at #1628 for inspiration and guidance on how it was done for the Suggest A Correction form.

PR #1638, opened 2026-01-30
Add devbox configuration

kayiwa:


tpendragon (changes requested):
@kayiwa We have to close a couple tickets to merge this, but I'll mark those sudden priority and we'll get to those ASAP.

Blockers: https://github.com/pulibrary/pulfalight/issues/1629

https://github.com/pulibrary/pulfalight/issues/1636

Some version of https://github.com/pulibrary/pulfalight/issues/1630

tpendragon (changes requested):
@kayiwa We got this done! We pulled your commit out here: https://github.com/pulibrary/pulfalight/pull/1645

The devbox stuff is nice, but our projects don't have all those extra commands, and I think our devs right now are used to those - I'm not sure we want to maintain those. Maybe just setup? I think it'd be good to have the devbox install script too, or maybe document how to install it using a centralized script.

Some formatting of that JSON would be good too, it'd be pretty hard to maintain as is.

https://github.com/pulibrary/figgy/blob/main/devbox.json is an example of one we have.

I also don't understand where some of the smaller README changes came from, did those feel better to you?

Thank you for working to make it easier to be in pulfalight!

PR #1692, opened 2026-07-15
Add quotes around interpolated password in config.yml

### LINKED ISSUE #1691: Psych::SyntaxError when setting up the project for the first time
When I follow the instructions in the README and `bundle exec rails s`, I get:

```
Psych::SyntaxError ((<unknown>): found character that cannot start any token while scanning for the next token at line 5 column 26):

config/initializers/pulfalight_config.rb:14:in 'Pulfalight.config_yaml'
config/initializers/pulfalight_config.rb:4:in 'Pulfalight.config'
config/initializers/pulfalight_config.rb:8:in 'Pulfalight.multiple_algorithms_enabled?'
[...]
```

Indeed, the value that would be interpolated into line 5 of config/config.yml begins with a somewhat unusual character.  Putting that interpolated value in quotes fixed it for me -- pr coming up.

sandbergja:
There is a chance that an ASpace password could include special characters, and should therefore be put in quotes when it is interpolated into a YAML file.

Closes #1691

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -2,7 +2,7 @@ defaults: &defaults
   external_universal_viewer_url: "https://figgy.princeton.edu/viewer"
   archivespace_url: <%= ENV["ASPACE_URL"] || "https://aspace.princeton.edu/staff/api" %>
   archivespace_user: <%= ENV["ASPACE_USER"] %>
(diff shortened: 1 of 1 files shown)


christinach (approved):
Thanks @sandbergja 🍄 !

hackartisan (approved):

Issue #1635, opened 2026-01-23
Make parent components sticky when navigating the ToC


faithc:
### Summary (Product Owner)
Users lose necessary context when navigating or browsing child components particularly when there are a fair number of child components associated with a parent component. This includes scenarios when users navigate to a particular component from the search results page.

Example: https://findingaids.princeton.edu/catalog/C1491_c1574. It's not clear and difficult for users to discern that this description falls under/is Correspondence. 

### Impact (Product Owner)
Context is key for users navigating finding aids/archival collections to understand the various materials included in a collection and their relationship to one another. Granular descriptions are associated with aggregate descriptions that provide additional information/context about the former.  

### Deadlines (Product Owner)
_Please include hard deadlines - events that can't be moved, people who can't work, etc._

### User Story (Product Owner) (Feature Request)
As a user (any persona), I want to be able to see how materials described at a granular level are related to those described at a higher, more aggregate level at all times to understand their relationship to one another in order to have a better understanding of materials described more granularly. 

### Priority recommendation (Product Owner)

- [ ] asap (add sudden priority justification below)
- [ ] within the next 3 weeks (add sudden priority justification below)
- [x] PO will prioritize

### Sudden Priority Justification (Product Owner)
_Say how staff’s work is being impacted and why there are no suitable workarounds. Add "Sudden Priority" and "dls-work-cycle" labels if you fill this out._

### Acceptance Criteria (Product Owner / Developers)

This work is complete when:

- [ ] (Product Owner Fill This Out)
- [ ] The UI implemented for this issue meets accessibility requirements
- [ ] New functionality is documented

sdellis:
If a user goes directly to a component (for example, through search result click) then all of the parent components must display above the component highlighted in the TOC and be sticky as they scroll. As soon as the next level is scrolled to, the new level would replace the old one in the the sticky stack of TOC elemets. 

We may want to explore other ways of solving this need (i.e., breadcrumbs or other way of showing hierarchical context.)

sdellis:
### Mockup Example 
[...]

PR #1696, opened 2026-07-15
[#1635] Make parent components sticky when navigating the ToC

sandbergja:
Add IntersectionObservers to each TOC leaf entry.  When one of the observers notices that it is displaying withing the ToC, it makes its parent's link sticky using the existing `.sticky-top` class.


christinach (approved):
Thanks @sandbergja 🍄 
Looks great! I also checked on qa.

Issue #505, opened 2020-11-19
Cart should close with escape


hackartisan:
This is standard modal behavior and would be very helpful for keyboard navigation

tpendragon:
Closed by #1699

PR #1699, opened 2026-07-15
When the focus is in the cart pressing escape should close the cart

christinach:
related to #505

This PR adds native [HTMLDialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog) element in the RequestCart template. Dialog supports the Escape keydown action by default. In our case we need to make sure that it also toggles the visibility in the store not just the native open and show dialog . 
When dialog opens it adds a top layer where all the elements are inter meaning not interactive. We want the user to be able to click outside of the cart and close the cart. For that reason we add a new click event on the dialog element where we're checking what the target is at the moment the user clicks. Anything outside of the cart is a dialog and if our check resolves to true we're closing it.
We've also tested the key navigation and it looks ok.

### DIFF
--- a/app/javascript/components/RequestCart.vue
+++ b/app/javascript/components/RequestCart.vue
@@ -1,9 +1,8 @@
 <template>
 
   <transition name="slide">
-
-  <div v-if="isVisible" :class="['request-cart']">
-
+    <dialog ref="dialog" class="request-cart">
+  
     <div class="panel">
       <table :class="['lux-data-table', 'fixed-header']">
 
@@ -164,8 +163,10 @@
       </div>
     </form>
 
-  </div>
+
+  </dialog>
   </transition>
+  
 </template>
 
 <script>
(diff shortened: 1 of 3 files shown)


tpendragon (changes requested):
Was it untenable to use the Dialog method linked in the issue? I think it'd take care of both your problems here, and might mean we have accessible modals going forward whenever that specification changes.

christinach:
> Was it untenable to use the Dialog method linked in the issue? I think it'd take care of both your problems here, and might mean we have accessible modals going forward whenever that specification changes.

I will check it! I forgot that it was referenced in the ticket.

app/javascript/components/RequestCart.vue
```diff
@@ -217,14 +216,30 @@ export default {
...
+          this.$refs.closeCart?.$el?.focus?.()
+        })
+      } else if (dialog.open) {
+        dialog.close()
       }
-    },
+    }
   },
   methods: {
+    handleDialogClick(event) {
```
tpendragon, line 233: Question: I think I heard you say this had to be added because clicking outside of the dialog wasn't working even with `showModal` and setting `closedby="any"`? When I've run into that before it turned out because it was a CSS thing - like the dialog was actually taking the whole screen, and it should just take a portion of it, so that there is an actual "outside." Was I understanding right? If so, might be worth looking at.

tpendragon (changes requested):
Just one question, this is looking great!

app/javascript/components/RequestCart.vue
christinach: I added some more explanation in the main comment of the pull request. The dialog does take the whole screen as a top layer. Do you want me to look if the dialog can take the size that the main panel of the cart takes? Do you think I could do this in a separate pull request? If no can take a look tomorrow.

christinach:
@ccleeton has reviewed the changes in the qa environment and said it looked ok to move forward. [slack thread](https://pulibrary.slack.com/archives/C01RXFRBHUL/p1784228721616129?thread_ts=1784149290.689679&cid=C01RXFRBHUL)

app/javascript/components/RequestCart.vue
tpendragon: Yes, I'm curious if the dialog can take the size that the main panel of the cart takes. I think `showModal` adds a little `::backdrop` pseudo-element that you can style that's "everything but the dialog", and clicking that backdrop closes it.

I think doing so would simplify this PR if it turns out to be possible, so I'm interested in seeing it here.

tpendragon: As requested, created https://github.com/pulibrary/pulfalight/issues/1714 as a follow-up to let this be merged.

tpendragon (approved):

PR #1702, opened 2026-07-15
Add breadcrumbs to show page

### LINKED ISSUE #1288: Add library location + breadcrumbs to component show page
See https://github.com/pulibrary/pulfalight/issues/1051

eliotjordan:
Closes #1288 

There is a lot of data duplicated in the TOC and this breadcrumb. We could make this only appear in mobile view - POs indicated that this was a primary use case. Or we could just leave it.

Screenshot 2026-07-15 at 4 34 38 PM

### DIFF
--- a/app/views/catalog/_component_summary.html.erb
+++ b/app/views/catalog/_component_summary.html.erb
@@ -1,4 +1,10 @@
 <div class="overview", data-component-id="<%= document.refs.first %>">
+  <% breadcrumbs = parents_to_links(document) %>
+  <% unless breadcrumbs.nil? %>
+    <div class='breadcrumb-links media'>
+      <span><%= parents_to_links(document) %></span>
+    </div>
+  <% end %>
   <div class="lux document-title">
     <h2><%= document.normalized_title %></h2>
     <div class="document-attributes">
--- a/app/helpers/pulfalight_helper.rb
+++ b/app/helpers/pulfalight_helper.rb
@@ -103,8 +103,7 @@ module PulfalightHelper
     end
 
     # Add library location to breadcrumbs
-    url = solr_document_path(document.collection_document.id) + "#access"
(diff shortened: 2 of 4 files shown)


hackartisan:
I wonder if it would look better without the folder icon. I'm not sure what that icon is supposed to be communicating, and it results in a bit of an indent.

app/views/catalog/_component_summary.html.erb
```diff
@@ -1,4 +1,11 @@
 <div class="overview", data-component-id="<%= document.refs.first %>">
+  <div class='breadcrumb-links media'>
+    <% breadcrumbs = parents_to_links(document) %>
+    <% unless breadcrumbs.nil? %>
+      <span class="media-body" aria-hidden="true"><%= blacklight_icon :folder, classes: 'al-folder-icon' %></span>
```
hackartisan: I'm not sure this would always be a folder? I think sometimes it's more like a document icon. I wonder if we really need it here, though.

hackartisan (changes requested):

eliotjordan:
@hackartisan It does look better without the folder.

Screenshot 2026-07-16 at 9 40 11 AM

app/views/catalog/_component_summary.html.erb
```diff
@@ -1,4 +1,10 @@
 <div class="overview", data-component-id="<%= document.refs.first %>">
+  <div class='breadcrumb-links media'>
```
hackartisan: I think the conditional should be around this div? Otherwise the div will be empty if breadcrumbs are nil?

hackartisan (changes requested):
@eliotjordan sorry, I just had one other question when looking more closely.

app/views/catalog/_component_summary.html.erb
eliotjordan: Good catch.

hackartisan (approved):

PR #1712, opened 2026-07-16
Enforce 100% coverage

### LINKED ISSUE #1709: Set minimum coverage
There is no minimum coverage value set in the SimpeCov configuration. We should set it to 100% like the rest of our projects.

eliotjordan:
Closes #1709

- Sets minimum coverage
- Removes unused AspaceStatus class file
- Stores complete set of coverage artifacts in CI
- Removes unused `combine_coverage.rb` script
- Removes some dead code
- Adds tests to boost coverage


sandbergja (approved):
Thanks, @eliotjordan !

PR #1713, opened 2026-07-16
Switch to happy-dom from jsdom

sandbergja:
Related to #505

We'd like to create a realistic test of the HTML dialog element.  However, this element is not implemented in jsdom (see https://github.com/jsdom/jsdom/issues/3294)

This PR switches to happy-dom, which does implement it, so we can write our test.


christinach (approved):
Thanks @sandbergja @christinach 🍄 🐬

Issue #1714, opened 2026-07-16
Rely on Dialog `showModal` logic for escaping the cart dialog


tpendragon:
### Summary (Product Owner)

As a follow up to https://github.com/pulibrary/pulfalight/pull/1699 we'd like to see if we can remove the click logic introduced in https://github.com/pulibrary/pulfalight/pull/1699/changes#diff-a7a642bc27b2bbf286f3a5e94a648c109a97c9e156861adf3723f66c296d9826R233 by making the dialog only take up the amount of space of the screen as it shows, without extra background. That should enable `showModal`'s implementation to handle closing by clicking out.

### Acceptance Criteria

- [ ] There's no click handler, but clicking outside of the cart closes it.

### First Step (Developers)

Look at the CSS, see if the test needs to change to click outside the dialog.

tpendragon:
I poked at this some.

`closedby="any"` makes it work, but `happydom` doesn't seem to support it, and Safari support seems limited. I had to update the chromium build being used in selenium to get it to run too (which we should probably do generally.) Given that background, I agree - it seems safer to explicitly bind the click.

While I was there I noticed the focus code though, I don't think we need that - it's one of the things showModal does. Added that as a PR: https://github.com/pulibrary/pulfalight/pull/1728

PR #1720, opened 2026-07-17
1714 cart click close

christinach:
related to #1714

I tried making the dialog size smaller and same as the panel using the pseudo attribute backtrop but didn't close the cart on a backtrop click. The online references I came across for the native dialog all described a fix for the backtrop click using javascript. 
The PR simplifies the click action: now it happens on the dialog and uses the native dialog close. That way it also lets the close handler to update the state TOGGLE_VISIBILITY.

The css to change the dialog background to transparent is just for display . The dialog's size remains the same as a top layer.

### DIFF
--- a/app/javascript/components/RequestCart.vue
+++ b/app/javascript/components/RequestCart.vue
@@ -1,7 +1,7 @@
 <template>
 
   <transition name="slide">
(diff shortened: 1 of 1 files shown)


sandbergja (approved):
🐬 Thanks, I like this simplification @christinach

PR #1724, opened 2026-07-20
Use fully qualified domain names for deploy

hackartisan:
You need these when deploying from a local machine (or doing deploy:rollback, which is what we were doing when we discovered this problem)


christinach (approved):
Thanks @hackartisan !

PR #1728, opened 2026-07-20
Let Dialog handle focus

tpendragon:
Dialog showModal handles focus, we don't need to specifically adjust it. Small thing I came across while poking at #1714


christinach (approved):
Thanks @tpendragon 🐻

PR #1783, opened 2026-07-27
Display bioghist notes in a modal

### LINKED ISSUE #1233: Display biographical/historical notes associated with subjects
Biographical or historical notes associated with name authorities should display for names that are associated as subjects in addition to those identified as creators. For example, see https://aspace.princeton.edu/staff/agents/agent_person/33165 (mss repo). See [public record](https://findingaids.princeton.edu/catalog/C0140_c03353) for item with which this name is associated.

tpendragon:
Closes #1233
Screenshot 2026-07-28 at 12 05 15 PM


app/helpers/pulfalight_helper.rb
```diff
@@ -151,8 +151,27 @@ def render_page_title
...
+      value = JSON.parse(value)
+      tag.span do
+        link_to_name_facet({ value: [value["label"]] }) + + bioghist_tag(value)
+      end
+    end.join("").html_safe
+  end
+
   private
 
+  def bioghist_tag(value)
```
eliotjordan, line 164: suggestio: I wonder if putting these two methods - or similar methods into the solr document is better. 

```
def name_bioghists; [..]; end

def name_bioghist(name); name_bioghists[name]; end
```

app/views/catalog/_show_upper_metadata_default.html.erb
```diff
@@ -38,9 +37,13 @@
       <% doc_presenter = document_presenter(document).with_field_group(metadata) %>
       <% generic_document_fields(metadata).each do |field_name, field| %>
         <% if generic_should_render_field?(metadata, document, field) %>
-          <%= render partial: "catalog/field", locals: { document: document, field: field, field_name: field_name, metadata: metadata, doc_presenter: doc_presenter } %>
+            <%= render partial: "catalog/field", locals: { document: document, field: field, field_name: field_name, metadata: metadata, doc_presenter: doc_presenter } %>
+            <% if field_name == "names_structured_ssm" %>
```
eliotjordan: suggestion: Instead of adding the logic here, you could add a helper method to the catalog controller that adds the links and markup which triggers the modal.

https://github.com/pulibrary/pulfalight/blob/main/app/controllers/catalog_controller.rb#L395

tpendragon: Aren't I doing this already in `structured_names_helper`? This turned out to just be for rendering the modal, I could probably move it elsewhere. But also, it felt kinda wrong to put HTML gen in a helper. 🤔

tpendragon:
Some adjustments I did after review:

* Pull out the modal partial and render it simpler.
* Add an accessor for the structure.
* Add a few comments.
* Style things a little nicer
* Make the "Historical Information" a button, not a link - modal triggers should be buttons.
* Replace bootstrap modals with a Dialog - I can't block another PR on not using Dialog and then use a bootstrap thing 😆

hackartisan (approved):
Nice updates

PR #1798, opened 2026-07-29
[WIP] Migrate to EntraID

### LINKED ISSUE #1796: Migrate from CAS to Entra-ID
### Summary or User Story

We've been asked to move our applications from CAS to Entra-ID by the end of the calendar year, barring any significant blockers.

[Documentation](https://github.com/pulibrary/pul-it-handbook/blob/main/services/entra-id.md) for how to add an Entra-ID app. We would be adding an OIDC connection for this app.

### Acceptance Criteria

- [ ] Login continues to work.
- [ ] OIDC accounts map to previously existing CAS accounts.

### First step

Read the docs, find the relevant library - probably https://github.com/omniauth/omniauth_openid_connect?

tpendragon:
I think this finally works. It took me a little while to get redirects working, and I had to adjust things like we did in pulmap so that you have to login via a POST button.

Some info:

You have to request `profile` in order to get `preferred_username`, and when you do what you get back is `netid@princeton.edu`. That's different from CAS, so in order to keep users without migrating, we gotta parse things a little differently.

The Entra ID tokens expire after a while. We need to figure out some process to rotate those. The more automatic the better. Right now I have this dev one set to 18 months.

Testing this was a nightmare to figure out, but I think I got it. I kept having this problem, specifically with sidekiq, where it would lose the path it redirected to in a feature spec. Turned out to be because of `app_host` and how cookies save with a host, I think.

Closes #1796


hackartisan:
What does an AD id look like?

tpendragon:
> What does an AD id look like?

I think the EntraID for me is "2U7A4RaKIwBm8AramwlTfj5_zASKnFLevIdHiucZrtc" and the OnPrem S-ID (which is local AD I think) is "S-1-5-21-1268611206-43474576-316617838-204483"

PR #1799, opened 2026-07-29
Visually improves grouping of results by collection

### LINKED ISSUE #1289: Implementation of Grouped Result Set Mockups
The Product Owners are satisfied with the mockup that intends to provide more clarity for grouped results in PulfaLight. 
Mockup References: #1050

Note: There is still dissatisfaction with the icons we use in the result sets. Resolutions/discussions of icons will be in a separate and future issue.

sdellis:
Closes #1289 

### Screenshot

Screenshot 2026-07-29 at 4 38 26 PM

### Screenshot Mobile

Screenshot 2026-07-29 at 4 40 58 PM


config/locales/arclight.en.yml
```diff
@@ -0,0 +1,99 @@
+---
```
eliotjordan: If you need to update individual values, you can add them individually to the [locales/en.yml file.](https://github.com/pulibrary/pulfalight/blob/main/config/locales/en.yml) rather than overriding the entire upstream `arclight.en.yml` config. 

I might remove `hello: "Hello world"` and the preceding comment paragraphs for clarity.

eliotjordan (changes requested):

config/locales/arclight.en.yml
sdellis: Done! Thank you!

eliotjordan (approved):

